PR #1, opened 2018-01-25
BWRift detailed state

groupsky:
This PR extends the data gathered in Bristle Woods Rift. Records buffs, chamber state and acolyte phases.

---

Haven't tested it extensively so I don't think it's ready for merge, but would like a general feedback. In particular:

1. Naming of the sub states
2. Is it ok to use booleans or should I convert to number/labels?

---

Things left to do:

- [ ] Handle Chronomaster dropping the hourglass
- [ ] Validate state while charging obelisk
- [ ] Validate state before having the hourglass

### DIFF
--- a/scripts/main.js
+++ b/scripts/main.js
@@ -255,7 +255,8 @@
 
         message = fixLGLocations(message, response, journal);
         message = getStage(message, response, journal);
-        message = fixTransitionMice(message, response, journal); // Must be after get stage to fix bad stages
+        message = getHuntDetails(message, response, journal);
+        message = fixTransitionMice(message, response, journal); // Must be after get stage and get details to fix bad stages
 
         if (!message || !message.location || !message.location.name) {
             window.console.log("MHHH: Missing Info (will try better next hunt)(2)");
@@ -447,11 +448,13 @@
                 break;
             case "Bristle Woods Rift":
                 if (message.mouse === "Absolute Acolyte" && message.caught === 1) {
-                    message.stage.stage_name = "Acolyte";
-                    message.stage.has_hourglass = true
-                    message.stage.chamber_status = 'closed'
-                    message.stage.obelisk_charged = true
-                    message.stage.acolyte_sand_drained = true
+                    message.stage = "Acolyte";
+                    if (message.hunt_details) {
+                        message.hunt_details.has_hourglass = true
+                        message.hunt_details.chamber_status = 'closed'
+                        message.hunt_details.obelisk_charged = true
+                        message.hunt_details.acolyte_sand_drained = true
+                    }
                 }
                 break;
             case "Burroughs Rift":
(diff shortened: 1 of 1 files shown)


DevJackSmith:
while this is a good idea, it is much harder to implement on the back end than it is on the extension. Extension just checks some logic and passes the values to the back end. (By the way on the back end message.stages is treated as a simple indexed array, so passing false, will just record stage "false", which could be anything by itself. So to answer your question we need labels/strings.)

Unfortunately this will really pollute (for a lack of better word) my attraction rate, looter and other tools. And make map helper tool just straight up useless. (This is due to there being at least 5 stage types for every combination of mouse+location+cheese). Assuming that this is rift and most people will only use rift cheese or sb+, this brings us to a lot of combinations (5*3*however many stages are in each stage type).

Which brings me to a hard choice (any advice appreciated):
 
1. not record all the details that I can and keep the simple high level tools
2. record all the details that I can, abandon the view tools
3. start changing and limiting the details of all the view tools (which I don't have time for, so this could be a year or more)
   (Maybe record all the details in a different way than stages, so my existing tools don't pick it up. This also requires a bit of planning and changing of db on the back end, but can be done. For example another table named "Details" with fields like hunt_id, name, value)

DevJackSmith:
and to answer your first question, I am not sure I fully understand. The variable names look ok to me, the actual labels/strings I would prefer to keep as close to how they appear in the game as possible. For example Acolyte or Entrance. But you already do that too.

groupsky:
> By the way on the back end message.stages is treated as a simple indexed array, so passing false, will just record stage "false", which could be anything by itself. So to answer your question we need labels/strings.

Good to know, I will refactor.

> Which brings me to a hard choice (any advice appreciated):
> 1. not record all the details that I can and keep the simple high level tools
> 2. record all the details that I can, abandon the view tools (this means completely relying on tools like Tsitu's to present my data)
> 3. a. start changing and limiting the details of all the view tools (which I don't have time for, so this could be a year or more)
>   b. (Maybe record all the details in a different way than stages, so my existing tools don't pick it up. This also requires a bit of planning and changing of db on the back end, but can be done. For example another table named "Details" with fields like hunt_id, name, value)

I think option 3b - recording in a new table, is the way to go. The existing tools will continue to work and gradually the new data dimensions can be integrated. Can try to help on the backend side if you need to.

DevJackSmith:
sounds good. If you can please create a new field on message object, lets name it message.hunt_details or something similar. I'll add it on the back end as a separate table.

groupsky:
Extracted as `message.hunt_details` key-value structure. If you prefer a list, I can refactor, but IMHO it's better with kv.

scripts/main.js
```diff
@@ -1091,6 +1098,36 @@
...
+        message.hunt_details = {}
+        var quest = response.user.quests.QuestRiftBristleWoods
+        for (var key in quest.status_effects) {
+            if (!quest.status_effects.hasOwnProperty(key)) continue
+            message.hunt_details['effect_'+key] = quest.status_effects[key] === 'active'
+        }
+        message.hunt_details.has_hourglass = quest.items.rift_hourglass_stat_item.quantity >= 1
+        message.hunt_details.chamber_status = quest.chamber_status
+        if (quest.chamber_name === 'Acolyte') {
+            message.hunt_details.obelisk_charged = quest.obelisk_percent === 100
```
DevJackSmith: I don't think this will work properly as the server response returns what is current, not what was during the hunt. So if the server returns 100, it's what it is after the hunt, not necessarily what was during the hunt. Not a stopper, just fyi.

DevJackSmith:
could you clarify what is the chamber_status?

scripts/main.js
groupsky: Correct, but in this case should make up for small offset in population distribution.

groupsky:
> could you clarify what is the chamber_status?

`open` is when portals are opened, `closed` otherwise

PR #3, opened 2018-02-27
Fix recording of M400 catches

groupsky:
When M400 is caught it disarms the cheese, but that breaks recording it's catch. Since it's only attracted to Fusion Fondue, this patch checks for the situation and manually sets the cheese for the hunt.

### DIFF
--- a/scripts/main.js
+++ b/scripts/main.js
@@ -248,6 +248,7 @@
         }
 
         message = getMainHuntInfo(message, response, journal);
+        message = fixM400Catch(message, response, journal);
         if (!message || !message.location || !message.location.name || !message.cheese.name || !message.trap.name || !message.base.name) {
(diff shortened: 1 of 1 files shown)


DevJackSmith:
great thinking, i like it a lot. if you don't mind i'll put it in with the rest of the transition mice fixes in fixTransitionMice, and just move the check for cheese a bit further. Great work :)

groupsky:
Moving the check further down the execution will not work due to [this check](https://github.com/DevJackSmith/mh-helper-extension/pull/3/files#diff-e8a233058211c77f51de535cd8ebb8d6R252) which drops the recording when there is no cheese.

DevJackSmith:
yes, I plan on moving just the cheese part of the check down, sorry if I wasn't clear.

PR #4, opened 2018-03-06
Track birthday 2018 area stage

groupsky:
Current year is tracked as stage, while boss status and furthest year are recorded as hunt details.


DevJackSmith:
So boss is a valid stage, as well as the stage after one catches the boss. Any idea what is the keys/values for the stage after one catches the boss?

DevJackSmith:
got report that once we repair year 2017, you stay forever in the boss stage, where all mice including the boss are attracted.

PR #5, opened 2018-04-04
Record Sunken City zone as hunt details

groupsky:
Current zone is recorded only if haven't changed zones. In that case just skip recording.


groupsky:
I was thinking about stages too, but it may be some time before it gets enough data to be really usable, so I think you may merge it as is, and in some time migrate the data as stages and switch the code.

PR #11, opened 2018-10-18
Moved "publish" string

AardWolf:
It looks like this is the only thing that actually moved. I don't know how to test it. The referenced item appears to refer to what is now the render_data.text in the journal entry and looks like the text one would see in the journal itself.

FWIW the journal.render.css_class contains "entry short active catchsuccessloot" for when I caught a mouse and got loot. Since nobody ever changes css class names those could be safe to check for catch/FTC/FTA - and whether to check for loot.


DevJackSmith:
thank you, but there are more places where i rely on journal.publish_data.attachment which no longer exists.

PR #14, opened 2018-11-21
Fixed login check alert spam

tsitu:
The game occasionally logs me out on both desktop & mobile, which causes [this](https://streamable.com/f1mof) to happen. I added a check that should limit the alerts to just 1 (instead of about 7).

### DIFF
--- a/scripts/main.js
+++ b/scripts/main.js
@@ -19,7 +19,7 @@
         if (ev.data.jacks_message === null) {
             return;
         }
-        if (typeof user.user_id === 'undefined') {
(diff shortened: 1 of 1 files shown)


scripts/main.js
```diff
@@ -19,7 +19,7 @@
         if (ev.data.jacks_message === null) {
```
DevJackSmith: found a better fix, this === should have been only ==

tehhowch: Probably the better expression then is `if (!ev.jacks_message)` if you want to indicate that **any** falsy value should enter this branch.

PR #15, opened 2019-02-06
Display crown count message after request resolves

tehhowch:
Refs https://github.com/DevJackSmith/mh-hunt-helper/issues/91

 - [x] Shows success / failure flash message, depending on whether the submission was sent out
    - Success toast includes the total crowns and submittee's name
 - [x] Error toasts are displayed longer
 - [x] Extends crown tracking possibility to players with a non-Facebook game account (having snuids which are `hg_` followed by a hash)
 - [x] Submits from the content script, where other extensions cannot block the request (e.g. Privacy Badger blocks the POST from main.js because it goes to `*.google.com/*`). 
    - Chrome still properly sets the DNT flag on the requests from the content script in my testing, in case users have enabled that
 - [x] On loading a `profile.php?snuid=` page, queries the badge data: users no longer must click the "Kings Crowns" tab first.
     - This greatly reduces the clicks needed for MHCC admins and any users accustomed to the HornTracker behavior (as the extension now emulates it).

Possible follow-ups:
 - [ ] Move all crown processing to content script, not just submission. Could perhaps simplify some code.

### DIFF
--- a/scripts/main.js
+++ b/scripts/main.js
@@ -60,6 +60,7 @@
             } else {
                 displayFlashMessage(ev.data.settings, "error", "There was an issue submitting crowns on the backend");
             }
+            return;
         }
 
     }, false);
@@ -224,7 +225,9 @@
         if (!settings.track_crowns) {
             return;
         }
-        let url_params = url.match(/snuid=([0-9]+)/);
+        // Traditional snuids are digit-only, but new snuids are `hg_` plus a hash, e.g.
+        //    hg_0ffb7add4e6e14d8e1147cb3f12fe84d
+        let url_params = url.match(/snuid=(\w+)/);
         if (!url_params || !Object.keys(xhr.responseJSON.mouse_data).length) {
             return;
         }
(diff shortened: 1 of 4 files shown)


DevJackSmith:
yes, extension is injected into the body of MH webpage. So when webpage is reloaded the extension has to be injected again, after page load. So i can't capture non-ajax unless i check the page on extension load, which could have fake data. For example i edit the page and then turn on the extension.

DevJackSmith:
i wonder if i can check for ajax without injecting my extension into the page... that could be a possibility.

tehhowch:
The latest commits add support for crown tracking of users with the "new" snuid pattern, and make it so the user does not have to explicitly load the "Kings Crowns" profile tab to track crowns. 

As mentioned in the Discord, there is currently a CORS issue when posting to the Apps Script endpoint from a profile page of a new snuid user. Traditional snuid profiles worked fine in my tests.

tehhowch:
The CORS issue is actually from an error in the script we are POSTing to - it doesn't return valid content for new snuid types, since an unhandled error is thrown when trying to insert `hg_{hash}` into a numeric column. So instead of redirecting to a valid output page (one that has `Access-Control-Allow-Origin` set), the script is served an error page that doesn't have any CORS headers set.

I checked the script project's dashboard (at https://script.google.com ) and noticed the error messages:
image

And also noticed that successful POSTs were 2-part, with an initial POST receiving HTTP 302 and then an ECHO, while the "failing" posts were just HTTP 200 to the target `scripts.google.com/macros/.../exec` URI

So the solution here is to update the Fusion Table and the Apps Script project (which also probably should not reject submissions that have 0 bronze crowns -- it should reject submissions where there are *no* crowns, or the timestamp is inaccurate by some delta).

scripts/main.js
```diff
@@ -54,11 +54,15 @@
...
 
+        // Crown submission results in either the boolean `false`, or the total submitted crowns.
         if (ev.data.jacks_message === 'crownSubmissionStatus') {
-            if (ev.data.submitted) {
-                displayFlashMessage(ev.data.settings, "success", "Thank you for submitting crowns!");
+            const counts = ev.data.submitted;
+            if (counts) {
+                displayFlashMessage(ev.data.settings, "success",
+                    `Submitted ${counts} crowns for `
+                    +  $('span.hunterInfoView-userName').text() + ".");
```
tsitu: Couldn't this be consolidated into the template literal?

```js
`Submitted ${counts} crowns for ${$('span.hunterInfoView-userName').text()}.`
```

tehhowch: Yep, it could. I originally was outputting more info (G/S/B individual) and broke it across lines to limit editor scrolling. Since the success flash is only shown for 1.5s it doesn't make sense to have a lot of detailed info in it.

PR #16, opened 2019-02-06
Use `string::includes` instead of `string::search !== -1`

tehhowch:
Probably this PR should be tested considerably; I wasn't able to experience all the journal entries to verify the associated correct behavior.
I left the `string::search` associated with Zokor stages since you're using a case insensitive `RegExp` there, and [`string::includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) accepts a string input.

I also wonder if we can use the new `mouse_type` property of `render_data` to more explicitly identify hunts from things like convertibles, etc.

This commit also includes a preliminary typedef for IDEs that support the JSDoc style; at some point these definitions could be improved and used elsewhere. I'm happy to remove it though if it isn't useful to have.

### DIFF
--- a/scripts/main.js
+++ b/scripts/main.js
@@ -245,15 +245,36 @@
        sendMessageToServer(map_intake_url, map);
     }
 
-    // Record successful hunt
+    /**
+     * Record a successful hunt.
+     * @param {XMLHttpRequest} xhr
+     */
     function recordHunt(xhr) {
+        /** @typedef {Object} JournalEntry
+         * @property {Object} publish_data
+         * @property {Object} wall_actions
+         * @property {Object} render_data
+         * @property {string} render_data.css_class The CSS classes associated with this journal entry
+         * @property {string} render_data.entry_date A time string (e.g. "10:44 am")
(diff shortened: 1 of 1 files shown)


DevJackSmith:
i'm not familiar with .some() yet, and you are using a different notation. Could you explain why this is better than what i have?

tehhowch:
`Array#some` is a self-shorting test for the callback to be true given the passed value. So as soon as the callback returns truthy for an element in the array, the expression resolves as `true`. I think of it as an `any` method, and it serves as the complement to the `Array#every` method (which returns `false` as soon as the callback returns falsy).

So rather than need to extend a regular expression, you only need to add the requisite string to the array. As for better, it's just different, and makes it very clear that we don't care where in the CSS / parent string the value is, just whether the term is present or absent.

DevJackSmith:
i'm old school, so my way seems much easier for me to maintain and troubleshoot, but who am i to say no to new shiny things... guess it's time i learn them... gonna be pain in the butt in a week after i forget what that is.

PR #17, opened 2019-02-07
Track LNY bonus luck via hunt_details

tehhowch:
- Restrict event mice check to only those done with event cheese (since Costumed ____ mice are not attracted to the stockpile's normal bait)
 - Track bonus luck from active lantern as a hunt detail
 - flag all hunts for which a lny quest exists (and thus a bonus luck amount may also exist)

### DIFF
--- a/scripts/main.js
+++ b/scripts/main.js
@@ -668,8 +668,12 @@
 
         // 2019 LNY costumed mice check
         let quest = response.user.quests.QuestLunarNewYear2019;
-        if (quest != null && quest.has_stockpile == "found" && !quest.mice.costumed_pig.includes("caught")) {
-            return "";
+        if (quest && quest.has_stockpile == "found" && !quest.mice.costumed_pig.includes("caught")) {
+            // Ignore event cheese hunts as the player is attracting the Costumed mice in a specific order.
+            let dumpling = quest.items.lunar_new_year_2017_cheese;
+            let nian = quest.items.lunar_new_year_2018_cheese;
+            if (dumpling.status === "active" || nian.status === "active")
+                return "";
         }
 
         return message;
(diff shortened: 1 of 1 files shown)


tehhowch:
Some corner cases to consider:

 - when the lantern reaches 500, does the height keep increasing? Or is it fixed at 500? If it's fixed at 500, we'll need a way to determine if we should use luck = 50 or 49.
 - should we ignore all LNY hunts where the lantern is inactive and the count of unlit items is 0, i.e. "we just ran out of candles and that's why the lantern is inactive"
  - There's no way to know how much bonus luck was applied if we consume the final candle, unless we obtain a valid pre-hunt snapshot and use that instead of the post-hunt response.

scripts/main.js
```diff
@@ -1271,6 +1275,40 @@
...
+                lny_luck = getLNYLuck(quest);
+            } else {
+                // The lantern is inactive. This may be an explicit choice, or because the user ran out of equipped candles.
+                let yellow = quest.items.lny_unlit_lantern_stat_item;
+                let red = quest.items.lny_unlit_lantern_2018_stat_item;
+                // If the user has both candles, we know the lantern state was an explicit
+                // choice. (Candles given via height rewards must be explicitly claimed.)
+                if (parseInt(yellow.quantity, 10) > 0 && parseInt(red.quantity, 10) > 0) {
+                    lny_luck = 0;
+                } else if (yellow.status === "active" || red.status === "active") {
```
tehhowch: The idea right here is that maybe one of the candles still reports if it was active even if the player ran out of it on the hunt. I'll be able to determine how the game responds in a few hours.

tehhowch:
Running out of candles generates a journal entry:

image

In `journal_markup`:
```json
{
  "render_data":{
    "image":[],
    "entry_id":340169,
    "mouse_type":false,
    "css_class":"entry short misc custom chesla_trap_trigger",
    "entry_date":"9:07 am",
    "environment":"Cantera Quarry",
    "social_link_data":{ 
      ...
    },
    "entry_timestamp":1549638440,
    "text":"I have run out of <b>Lunar Lantern Candles<\/b>! I will no longer benefit from the magic of the Pig Lunar Lantern!"
  },
  "publish_data":{
    "type":"journal",
    "params":{
      "user_id":#######,
      "journal_id":340169
    },
    "quote":null,
    "params_string":"user_id=#######&journal_id=340169"
  },
}
```

PR #19, opened 2019-02-14
Refactor background.js code

tehhowch:
- Update comments
 - Add some error handling re: messaging unavailable tabs & checking `chrome.runtime.lastError`
 - Simplifies notification logic
 - Improves timer accuracy by rounding to nearest minute, rather than flooring unconditionally. e.g. 6:59 is now reported as "7m" instead of "6m"

### DIFF
--- a/scripts/background.js
+++ b/scripts/background.js
@@ -138,12 +138,16 @@ function icon_timer_updateBadge(tab_id, settings) {
             if (settings.icon_timer) {
                 chrome.browserAction.setBadgeBackgroundColor({color: '#222'});
                 response = response.replace(':', '');
-                let response_int = parseInt(response);
+                let response_int = parseInt(response, 10);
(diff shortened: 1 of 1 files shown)


scripts/background.js
```diff
@@ -1,10 +1,16 @@
...
+
 // Update check
 chrome.runtime.onUpdateAvailable.addListener(details => {
-    console.log("MHHH: updating to version " + details.version);
+    console.log(`MHHH: updating to version ${details.version}`);
     chrome.runtime.reload();
 });
 
-var time_interval = 7200 * 1000; // seconds * 1000
+// TODO: Do we need to request the update at all? Chrome auto-checks automatically.
```
DevJackSmith: we do not need to request update at all, but i'm not sure how fast Opera works. I did this originally to make sure the addon updates asap. We can take this code out. good catch.

PR #20, opened 2019-02-14
Refactor options.js code

tehhowch:
- Update comments
 - Inline event handler definitions
 - Use dot notation per JSHint suggestion

### DIFF
--- a/options.html
+++ b/options.html
@@ -122,12 +122,10 @@
                 </div>
             </div>
         </div>
-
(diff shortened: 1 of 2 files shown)


DevJackSmith:
thank you. looks good, I just need to test

tehhowch:
Updated the HTML to always show the "Save" button:

image


I'm open to css / class tweaks that will add padding around the button, if so desired.

PR #23, opened 2019-02-22
Populate hunt data using the pre-hunt user object

tehhowch:
Ditches the use of the in-response user object for building the hunt record that will be submitted. This resolves a host of issues with the established order, since we no longer are unsure about the exact setup used for the hunt.
Closes #17 
Closes #21 
Closes #22 
Closes https://github.com/DevJackSmith/mh-hunt-helper/issues/70
Refs https://github.com/DevJackSmith/mh-hunt-helper/issues/95 (Stage data from past extension versions will need cleaning / removal)
Refs https://github.com/DevJackSmith/mh-hunt-helper/issues/43 (Stage data from past extension versions will need cleaning / removal)
Closes https://github.com/DevJackSmith/mh-hunt-helper/issues/5
Closes https://github.com/DevJackSmith/mh-hunt-helper/issues/1

 - [x] **Acquire up-to-date user object prior to sounding the horn.**
This is done by intercepting the call to `activeturn.php` and delaying it until after we can query the title page. Per **cardboardshark**, supplying the page parameter `force=true` calculates any trap checks that should have occurred. I have verified that this is the case, and that the pre-hunt user object reflects changes made from the app, a different tab, and a different browser.
 - [x] **Track Grub Salt when hunting King Grub / King Scarab.**
This doesn't have any effect on the attraction pool, but for CRE folks this is a must-have. These salt levels are robust to running out of scent charms, catching KG/KS, and so on.
 - [x] **Track ZT progress**
Adds `hunt_details` for progress through Mystic & Technic pieces, and a boolean for whether the CM was attractable on the hunt. 0-7: pawns, 8-9 add knights, 10-11 add bishops, 12-13 add rooks, 14 add queen, 15 add king, and 16 means only pawn + rook (& cm) -- plus whatever is open on the other side.
 - **Convert from a `switch` chain to a "dispatcher" pattern for adding stage & hunt details.**
All stage functions have the same signature and as such we can make understanding the behavior very clear: "if this location has a stage function, we need to use it". (I'm also pretty sure that a hash lookup into an existing object is faster than evaluating the switch chain, but I'm not going to profile it). The only downside I see is that if multiple locations share the same stage function, we have to write the stage function name for each location, instead of using `case` fallthrough.
 - **Convert from functions that both modify an input parameter and return the modified input parameter to only modify the input parameter.**
This lets us use `const` to declare the `message`, providing assurances about its possible state (either `null`, if it could not be created, else always an object).
   - To signal a fatal error, the previous design was to assign a falsy value to `message`. Since `message` is protected from reassignment, we can signal to ignore the hunt by throwing an error (reserved for severe cases) or assigning a `null` location (for known hunts we don't care to record, for whatever reason). I don't expect many of the cases with `throw` statements to be reached at all. In the future we should work out a method to log these errors where the dev team is able to observe and respond to them.

 - There's a fair number of new logging statements, but these are gated behind a default-`false` user setting. While I and others evaluate the behavior, these are useful messages to have. Even after we're confident in the behavior, it could be useful to keep the log statements. In debug_logging mode, an exhaustive diff of the pre & post user objects is computed and printed in the console. For example, if a cheese / charm was disarmed or ran out, the location changed, quest / minigame progress occurred, or actual setup values like luck or power change, these will all be included in the diff.

With regards to verifying we obtained the actual pre-hunt object, two keys are required to change from pre-active-hunt to post-active-hunt: `num_active_turns` and `next_activeturn_seconds`. All other properties of the `user` object may change or not, depending on the hunt result, and not necessarily in logically useful manners. E.g. weapon or base components may become broken, cheese or charms may run out, be pillaged, or self-disarm, the location may change, and so on.  
 - It is possible, though unlikely, that a user attempts to simultaneously sound the horn and change setup through the use of multiple MH instances. In the future we could block hunt submissions in which the elapsed time between horn click and received hunt response exceeds some elapsed time, if desired. We could additionally subscribe to on-page events that trigger on user setup changes, e.g. to then abort any pending hunt submissions, but these events would not reflect changes made via other MH instances.


tehhowch:
I wonder re: charm-specifics (like Artisan, Prospectors, etc) if those should be ignored (they aren't specifically the *stage*) and instead await charm-specific views/interactions (e.g. https://github.com/DevJackSmith/mh-hunt-helper/issues/30, https://github.com/DevJackSmith/mh-hunt-helper/issues/94, https://github.com/DevJackSmith/mh-map-helper/issues/29).

My thought is that, while we can make the stage a "combo" field that reflects both charm & environment state, we perhaps should only do that in the interpretations of the stored data, and not the stored data itself.

DevJackSmith:
@tehhowch @tsitu I definitely agree that we should hide all the console log messages behind a flag. I'll try to look over and merge this one. This is a huge PR. Please keep them as little as possible with a single purpose. For example 20 lines changed maximum. Or if changing whole file spacing/format, do that and nothing else. That will help us review and approve them much faster. Piling on more stuff will only slow down our process. Also i'd suggest not adding anything after approval, but instead making another PR if possible.

DevJackSmith:
@tehhowch @tsitu did anyone test this?

tehhowch:
@DevJackSmith for PRs that change the architecture, they're generally always going to be large. I think if I were to remove the JSDoc function annotations that there would be a net decrease, so the diff isn't quite as large as the numbers make it seem.

As far as testing this, yes, I was using this branch as a developer extension while making the changes and updates. If you want, you could configure the server to route specific extension version hunts to a dev table, and then before merging this, check that the dev table successfully acquired hunts from all locations & stages. (PM tsitu and me the needed magic extension version to make that happen).

tsitu:
In terms of testing, I went on hunts in a few separate locations (with sendMessageToServer commented out) and inspected the console outputs for proper pre/post diffing behavior, as well as the final message object that would be sent to your server. I didn't travel to every location that had its logic changed, but I did read through each individual commit to make sure they were kosher.

I agree that this is a super hefty PR, but the fact that it's essentially a complete rewrite of the data submission process at least partly justifies it. If you'd prefer to do extensive testing with field data before deploying, we could adopt a debug extension version number and enlist some trusted users on Discord to checkout/install the extension from this PR to gather more data points.

tehhowch:
The logs are now gated behind a `debug_logging` setting

DevJackSmith:
Thanks for explaining. i like the idea about the debug tables. It will take me a while to review this thoroughly and approve this, but i'll try to get through this. Please stop adding changes to this PR. Even if we are switching to new way of checking for things, we can do it incrementally. The debug flag could also go into it's own pr. Next pr longer than 50 lines changed with multiple purposes will be auto-rejected. Think of PRs as a single purpose. Nothing extra. Extra feature = another PR. I'd rather have 200 PRs than 1 PR that will take me weeks to go through.

DevJackSmith:
i understand that you enjoy working on this, and i appreciate your hard work. I just can't dedicate more than half hour at a time to this project right now. So big PRs will take much much longer to review, test, merge, and release.

scripts/main.js
```diff
@@ -388,58 +472,133 @@
...
-            let prop_name = component.prop + '_name';
-            let prop_id = component.prop + '_item_id';
-            if (!user_resp[prop_name]) return;
-            message[component.message_field] = {
-                id: user_resp[prop_id],
-                name: user_resp[prop_name].replace(component.replacer, '')
+            const prop_name = `${component.prop}_name`;
+            const prop_id = `${component.prop}_item_id`;
+            const item_name = user[prop_name];
+            message[component.message_field] = (!item_name) ? {} : {
```
tehhowch: Note: For truly charmless hunts, this will create a field `charm: {}`

I'm not sure how that field is handled by Jack's server; if it's problematic this line should be updated

DevJackSmith: good note, thank you. I believe i will catch it just fine: https://github.com/DevJackSmith/mh-hunt-helper/blob/36d51a8a8ae6a9596119ee79c61079e880e241b1/Server/intake.php#L111 But i'd keep it the same type in future if possible: if the value is usually a string, then "" if it's missing. If the value is usually and object, then {} if it's missing.

DevJackSmith:
full path to mh.com might break fb. I don't have an fb account to test :(

tehhowch:
It was originally that: 

https://github.com/DevJackSmith/mh-helper-extension/blob/3150eeb2de82c73dc63118c222f5b49c5f0386ea/scripts/main.js#L1458

In the merge commit I shortened it to a relative path. Testing showed that the relative path required clicking on the profile tab to get the crown submission working again on profile load, so I reverted that url shortening.

DevJackSmith:
one more question, won't throwing errors quit the whole script? that would stop it from listening for ajax and not record hunts after that?

tehhowch:
The extension initiates with an IIFE, and all subsequent actions are the result of an invoked event listener. Unless throwing an error unsubscribes the event listener, then the next event should be handled as normal. 

The only errors that would stop all processing are those that occur while the  IIFE is executing, right as the page loads.

PR #25, opened 2019-03-05
Capture stage for MH Birthday 2019

tehhowch:
- Remove "Mysterious Anomaly" code as it is no longer relevant.

This implementation drops all birthday location hunts not done in one of the 4 rooms.
It's possible we should perform a "cheese check" since the location details / Larry tip strongly suggests that the stage only matters when `quest.items.coggy_colby_cheese.status === "active"` is true (i.e. non event cheese shares same pool, excepting SB+'s extra mouse.

I have the associated methods available for after prehunt goes live but won't add that until required via merge conflict.

### DIFF
--- a/scripts/main.js
+++ b/scripts/main.js
@@ -1256,12 +1256,12 @@
         const quest = response.user.quests.QuestBirthday2019;
         if (message.mouse === "Vincent, The Magnificent" || quest.factory_atts.boss_warning) {
             message.stage = "Boss";
-        } else {
+        } else if (quest.items.coggy_colby_cheese.status === "active") {
             message.stage = (({
-                "pumping_room":           "Pump",
-                "mixing_room":            "Mixing",
-                "break_room":             "Break",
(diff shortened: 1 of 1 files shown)


tehhowch:
No need for `hunt_details` that track the upgrade status of the Break Room: [![][1]][1]

As far as the `QuestBirthday2019` data goes, there's only a few properties of use:

`bait_warning`: bool/null, whether the user is using coggy (`null`) or other (`true`)
`factory_atts.can_claim`: bool/null, whether a cheese crate is ready to be recovered. probably no influence on attraction or loot
`factory_atts.boss_warning`: bool/null, unclear if this means *only* Vincent is attractable, or whether Vincent is now in the pool
`items.coggy_colby_cheese.status`: `"active"`, `"disabled"`, `""`, depending on if the event cheese is armed, not owned, or just not armed.

[1]: https://media.discordapp.net/attachments/552486487837507584/552558768035069952/image0.png?width=400&height=300

PR #26, opened 2019-03-07
Fixed inconsistent Auto-Loader bookmarklet injection

tsitu:
Moved "Display Tsitu's Loader" logic inside the load event handler for `main.js` to prevent it from running before `main.js` is executed, which leaves the message event unhandled.


DevJackSmith:
getSettings function should do one thing and one thing only. Just like all the other functions should.

PR #27, opened 2019-03-07
Switch to "no-cors" request to enable crown submissions from FF

tehhowch:
Refs DevJackSmith/mh-hunt-helper#98

I was unable to resolve the NetworkError with either `window.fetch`, `XMLHttpRequest`, adding `"*://script.google.com/*"` and `"webRequest"` to the extension manifests' `permissions`, and combinations thereof. 

I verified that, in `"no-cors"` mode, both Firefox and Chrome do actually make the submission to the Apps Script endpoint with my normal privacy-focused extensions both disabled and enabled, so I think this is OK.

I'm open to some other solution that makes it possible to know whether the request was even issued, but until then we'll have to settle for blind-firing the request. If we eventually add a way to query the user's crown history, then we could test via that method whether a new crown snapshot was recorded while still maintaining the "no-cors" submission request.

This PR also adds a background script listener for warning, log, or error messages coming from the extension. Previously there was no way to log anything on the background console, which is persistent across page reloads (unlike the default browser console).

### DIFF
--- a/scripts/background.js
+++ b/scripts/background.js
@@ -145,8 +145,9 @@ function icon_timer_updateBadge(tab_id, settings) {
                     if (response_int > 59) {
                         let minutes = Math.floor(response_int / 100);
                         const seconds = response_int % 100;
-                        if (seconds > 30)
+                        if (seconds > 30) {
                             ++minutes;
+                        }
                         response = minutes + 'm';
                     } else {
                         response = response_int + 's';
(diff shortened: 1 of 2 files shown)


scripts/background.js
```diff
@@ -159,3 +160,18 @@ function icon_timer_updateBadge(tab_id, settings) {
         }
     });
 }
+
+// Handle messages sent by the extension to the runtime.
+chrome.runtime.onMessage.addListener((msg, sender, sendResponse) => {
```
DevJackSmith: This is not just extension, this is all messages going on in the browser, no? Why are we even adding this in this PR? this should have been a 1 line PR changing "cors" to "no-cors" (maybe 3 lines to make the call opaque).

tehhowch: It's extension specific. The Firefox add-on debugger (at least for temporary) does not restrict what it displays to the debugged extension/add-on.

The message listener is added here because extension messaging is used by crown submission to report debugging level errors, and previously was not added - sent messages had no listener.

DevJackSmith:
> Previously there was no way to log anything on the background console, which is persistent across page reloads (unlike the default browser console).

console.log? why would we want something else to persist? btw in browser console you can choose persistent mode.

Much better than the kill-me-now PR, but this should've been it's own PR though. That way I would've already merged the no-cors one. Just saying. Anyway, thank you for your help, as always i'm grateful, (a bit confused, but grateful).

PR #28, opened 2019-03-10
Track Fort Rox upgrades & autocatch ability

tehhowch:
With these details, we can determine catch rates for the auto-catch mice, and compare catch rates before & after weakening, etc.

It's possible we could inspect other journal entries that came along in the activeturn response to determine if an autocatch ability activated on the given hunt, but that is not added here.

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1367,6 +1367,7 @@
     /** @type {Object <string, Function>} */
     const location_huntdetails_lookup = {
         "Bristle Woods Rift": addBristleWoodsRiftHuntDetails,
+        "Fort Rox": addFortRoxHuntDetails,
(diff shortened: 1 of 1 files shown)


tsitu:
I don't think these particular details are in urgent need of being tracked.

We know that Level 1 upgrades cut Weremice/Critter powers in half, and there is strong anecdotal data to suggest that Level 2 upgrades have a 50% insta-catch proc rate (as in `0.5 * (1 - CR)`).

However, it may be worthwhile to verify that Ballista/Cannon rate, as well as derive the (currently unknown but probably also 50%) Mage Tower rate.

tehhowch:
After the event ends, I'll spend a little time in FR to see what journal parsing we'd need to do to find out if the hunt was an autocatch.

PR #29, opened 2019-03-10
Track WWR rage levels for Lactrodectus usage

tehhowch:
The current WWR stage information is a string-based categorical variable for the 3 rage states. This PR adds a numerical value for each level, a true/false boolean for whether the attraction criteria is met (all >= 25), and the total rage sum.

It's possible the sum and boolean are unnecessary, as that information can conceivably be derived from the numerical levels in a SQL query.

I don't think knowing the numeric levels is important for other cheeses, which is why the cheese check is performed.

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1473,7 +1473,7 @@
     }
 
     /**
-     * For MBW attractions, record the rage state of each zone.
+     * For Lactrodectus hunts, if MBW can be attracted (and is not guaranteed), record the rage state.
      * @param {Object <string, any>} message The message to be sent.
      * @param {Object <string, any>} user The user state object, when the hunt was invoked (pre-hunt).
      * @param {Object <string, any>} user_post The user state object, after the hunt.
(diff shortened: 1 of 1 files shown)


tsitu:
Is the motivation here to track LL populations within the [75, 150) total rage range?

If so, it seems advisable not to attach details when total rage < 75 (MBW = 0%) and when total range is 150 (MBW = 100%)

tehhowch:
Yeah, hunting MBW at lower rage isn't a high frequency occasion. I think the original idea was that there could be some benefit to going for MBW at lower rage, in order to get more MBW / hunt on average. 

I have a scheduled query that runs against HT data (12k MBW attractions) and makes this chart:



A probably safe estimate is that the attraction rate increases linearly from 50% at 75 rage to 100% at 150 rage.

The query variables, where `<value>` is stepped 75-150:
```json
"vars": {
  "Cheese": { "100": { "exclude": false }},
  "Location": { "334": { "exclude": false }},
  "Total Rage Level": { "yada": {"min": <value>,"max": <value>, "exclude": false }},
  "Mouse": {
    "305": { "exclude": true},
    "271": { "exclude": true},
    "277": { "exclude": true},
    "278": { "exclude": true},
    "279": { "exclude": true},
    "280": { "exclude": true}
  },
  "Deep Lagoon Level":{ "d1": { "min": 25, "max": 50, "exclude": false}},
  "Crazed Clearing Level": { "d2": {"min": 25, "max": 50, "exclude": false}},
  "Gigantic Gnarled Tree Level": {"d3":{ "min": 25, "max": 50, "exclude": false}}
}
```

PR #31, opened 2019-03-10
Track Wanted Poster name & boss readiness

tehhowch:
This PR adds hunt details to Claw Shot City that can help us determine the attraction rate of specific ringleaders after defeating the rest of the gang.

I want to check the quest data associated with the boss state before merging this, as it's possible the CSC quest has some information.


tehhowch:
Check completed, nothing more we can really add to CSC data unless we defer hunt submission until after we receive data from a POST to `relichunter.php`, and even then we don't know we needed that data until after the hunt is done -- any POST would give us the post-hunt map mice remaining, and not reflect the mice available for the hunt.

PR #34, opened 2019-03-11
Submit crown counts via the background page

tehhowch:
This avoids invoking the NetworkError on Firefox, while still maintaining the isolation from other user extensions that was blocking submissions from occurring when submitted from the injected script.

Closes https://github.com/DevJackSmith/mh-hunt-helper/issues/98


DevJackSmith:
why did it try to build???

PR #40, opened 2019-04-03
Add Fiery Warpath "hunt details"

tehhowch:
- Report remaining mouse counts per wave, & if bosses are invincible.
 - Report `morale` / `support_morale` fields, for Warpath Thrasher computations
 - Report streak count, type, and whether the hunt extended it

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1532,7 +1532,7 @@
         const attrs = user.viewing_atts.desert_warpath;
         const fw = {};
         if ([1, 2, 3].includes(parseInt(attrs.wave, 10))) {
-            const asType = name => name.replace(/desert|_weak|_epic|_strong/g, "");
+            const asType = name => name.replace(/desert_|_weak|_epic|_strong/g, "");
 
             if (attrs.streak_quantity > 0) {
                 fw.streak = {
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1379,6 +1379,7 @@
         "Bristle Woods Rift": addBristleWoodsRiftHuntDetails,
         "Harbour": addHarbourHuntDetails,
         "Fort Rox": addFortRoxHuntDetails,
+        "Fiery Warpath": addFieryWarpathHuntDetails,
         "Sand Crypts": addSandCryptsHuntDetails,
         "Whisker Woods Rift": addWhiskerWoodsRiftHuntDetails,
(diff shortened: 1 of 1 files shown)


tsitu:
Quick Q's:

- Do Anniversary traps have an insta-catch chance on Warmonger & Artillery Commander with Wardens up? If so, would that affect the invincibility logic?
- What insights are you hoping to gain from recording streak count, type, and extended?

tehhowch:
1. I believe they do, so knowing if the boss should have been invincible helps verify the proc rate (not that many will use that anniv. traps here)
2. Several things:
  - Gargantuamouse attraction rate vs. streak
  - on/off streak attraction. There are at least two ways for the charms to bias the population, one being adding "pseudomice" and another being a flat proportion for on/off. The "streak extended" and type fields offer (I think) a great simplification of the otherwise-needed SQL to find instances where the right charm did or did not work

tsitu:
I don't quite follow what you mean by the charms adding "pseudomice".

Also, couldn't the flat proportion be the same no matter whether you're on or off a streak? For instance, suppose Super Warrior charms give a 50% attraction boost to Warrior-group mice. Wouldn't that 50% likely be the same whether or not you're on a streak of Warriors?

tehhowch:
By pseudo mice I mean that the population proportion is biased by some amount. Say if you were 3/3/9, arming an Archer charm might double the amount of archers, making it seem like 3/3/18. Arming a Super Archer might quadruple it, so it seems like 3/3/36.

Probably this is not the mechanism (charms anecdotally feel more effective than this scheme would suggest for wave starts and type eradications) but it would be nice to be definitive.

You are correct that a flat proportion is streak independent. I'm taking a shortcut and using HG's computation of whether the caught mouse is of the correct type to extend the streak. Changing the field to be streak independent, e.g. "targeted_was_attracted", means we have to do our own charm vs mouse comparison upon catch in order to set the flag correctly.

I envision the current SQL query to be something like, "select count() from hunts where locationid = # and details.streak.type = 'archer' group by cheeseid, charmid, details.num_archer, details.streak.increased_on_hunt".
This could certainly be doable with the other pattern: "select count() from hunts where locationid =# and details.target_type = 'archer' group by ..., details.num_archer, details.targeted_was_attracted".

To summarize I think there are definitely multiple ways to get at this same information. I went for streak extension because it meant simpler code :)

PR #59, opened 2019-04-10
Add SEH hunt details

tehhowch:
- Eggscavator Charge Level
 - Whether egg loot is doubled (useful to know for those comparing charge cycles)

IIRC the charge egg dropped from the hunt was from the higher of the two states, pre- vs post-hunt. But I can't be certain, so it's tracking both. It's also possible this behavior changes/changed, and without knowing both pre/post, data sifters may draw the wrong conclusion. (This is also why I'm not bothering with the low/med/high; we could add that or back-calculate & update after we know more.)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1438,16 +1438,14 @@
      */
     function addEggHuntDetails(message, user, user_post, hunt) {
         const quest = getActiveSEHQuest(user.quests);
-        if (quest) {
+        const post_quest = getActiveSEHQuest(user_post.quests);
+        if (quest && post_quest) {
             message.hunt_details = Object.assign(message.hunt_details || {}, {
                 is_egg_hunt: true,
                 egg_charge_pre: parseInt(quest.charge_quantity, 10),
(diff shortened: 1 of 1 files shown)


tehhowch:
Just confirmed it is still in effect this event: Beginning with 17 & ending with 18 charge yields High Charge eggs, beginning with 18 & ending with 17 gives High Charge eggs.

src/scripts/main.js
```diff
@@ -1403,11 +1403,34 @@
...
+     */
+    function addEggHuntDetails(message, user, user_post, hunt) {
+        const quest = getActiveSEHQuest(user.quests);
+        if (quest) {
+            message.hunt_details = Object.assign(message.hunt_details || {}, {
+                is_egg_hunt: true,
+                egg_charge_pre: parseInt(quest.charge_quantity, 10),
+                can_double_eggs: (quest.charge_doubler === "active"),
+            });
+            const post_quest = getActiveSEHQuest(user_post.quests);
```
tsitu: Perhaps move the `post_quest` check up a few lines so that details are only assigned when both pre/post are available?

PR #67, opened 2019-09-18
Fix selector for current profile's name

tehhowch:
The element is now `<span class="friendsPage-friendRow-titleBar-name" title="George Costanza">George Costanza</span>`


tsitu:
LGTM

PR #69, opened 2019-10-11
Add stage and hunt details for Valour Rift

groupsky:
Records each floor as separate stage

TODO: ultimate umbra states are not checked.

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1419,7 +1419,7 @@
                     stageName = "Floors 9-15";
                 } else if (floor >= 17 && floor <= 23) {
                     stageName = "Floors 17-23";
-                } else {
+                } else if (floor >= 25) {
                     stageName = "Floors 25-31+";
                 }
 
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -751,6 +751,7 @@
         "Sunken City": addSunkenCityStage,
         "Toxic Spill": addToxicSpillStage,
         "Twisted Garden": addGardenStage,
+        "Valour Rift": addValourRiftStage,
         "Whisker Woods Rift": addWhiskerWoodsRiftStage,
         "Zokor": addZokorStage,
     };
@@ -1396,6 +1397,30 @@
         }
     }
 
+    /**
+     * Report the tower floor and prestige
+     * @param {Object <string, any>} message The message to be sent.
+     * @param {Object <string, any>} user The user state object, when the hunt was invoked (pre-hunt).
+     * @param {Object <string, any>} user_post The user state object, after the hunt.
+     * @param {Object <string, any>} hunt The journal entry corresponding to the active hunt.
+     */
+    function addValourRiftStage(message, user, user_post, hunt) {
+        const attrs = user.environment_atts;
+        switch (attrs.state) {
+            case 'tower':
+                if (attrs.active_augmentations.tu) {
+                    // Umbra Floor #
+                    message.stage = "Umbra Floor " + attrs.floor;
+                } else {
+                    // Floor #
+                    message.stage = "Floor " + attrs.floor;
+                }
+                break;
+            default:
+                message.stage = "Outside";
+        }
+    }
+    
     /** @type {Object <string, Function>} */
     const location_huntdetails_lookup = {
         "Bristle Woods Rift": addBristleWoodsRiftHuntDetails,
@@ -1404,6 +1429,7 @@
         "Fort Rox": addFortRoxHuntDetails,
         "Harbour": addHarbourHuntDetails,
         "Sand Crypts": addSandCryptsHuntDetails,
+        "Valour Rift": addValourRiftHuntDetails,
         "Whisker Woods Rift": addWhiskerWoodsRiftHuntDetails,
         "Zokor": addZokorHuntDetails,
         "Zugzwang's Tower": addZugzwangsTowerHuntDetails
@@ -1734,6 +1760,36 @@
         };
         zt.cm_available = (zt.technic === 16 || zt.mystic === 16) && message.cheese.id === 371;
         message.hunt_details = zt;
+    }
+    
+    /**
+     * Report the active buffs and floor details
+     * @param {Object <string, any>} message The message to be sent.
+     * @param {Object <string, any>} user The user state object, when the hunt was invoked (pre-hunt).
+     * @param {Object <string, any>} user_post The user state object, after the hunt.
+     * @param {Object <string, any>} hunt The journal entry corresponding to the active hunt.
+     */
+    function addValourRiftStage(message, user, user_post, hunt) {
+        const attrs = user.environment_atts;
+        message.hunt_details = {
+            // Elexir Rain
+            er: attrs.active_augmentations.er,
+            // Sigil Hunter
+            sh: attrs.active_augmentations.hr,
+            // Secret Research
+            sr: attrs.active_augmentations.sr,
+            // Super Siphon
+            ss: attrs.active_augmentations.ss,
+            // String Stepping
+            st: attrs.active_augmentations.sste,
+            // Ultimate Umbra
+            uu: attrs.active_augmentations.tu,
+            // Prestige is incremented after each eclipse in the current run. Starts at 1
+            prestige: attrs.prestige,
+            // Floor type is 1-8, or 7. 
+            // TODO: check eclipse state
+            floor_type: attrs.floor_type,
+        };
     }
 
     /**
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1407,26 +1407,29 @@
     function addValourRiftStage(message, user, user_post, hunt) {
         const attrs = user.environment_atts;
         switch (attrs.state) {
-            case 'tower':
-                var floor, prefix;
-                if (attrs.floor % 8) {
-                    floor = 'Eclipse';
-                } else if (attrs.floor < 25) {
-                    floor = attrs.floor
+            case "tower":
+                var floor = attrs.floor;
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1735,6 +1761,36 @@
...
+            sr: attrs.active_augmentations.sr,
+            // Super Siphon
+            ss: attrs.active_augmentations.ss,
+            // String Stepping
+            st: attrs.active_augmentations.sste,
+            // Ultimate Umbra
+            uu: attrs.active_augmentations.tu,
+            // Prestige is incremented after each eclipse in the current run. Starts at 1
+            prestige: attrs.prestige,
+            // Floor type is 1-8, or 7.
```
tehhowch: ? 7 is within 1-8

src/scripts/main.js
```diff
@@ -1735,6 +1761,36 @@
...
     }
+    
+    /**
+     * Report the active buffs and floor details
+     * @param {Object <string, any>} message The message to be sent.
+     * @param {Object <string, any>} user The user state object, when the hunt was invoked (pre-hunt).
+     * @param {Object <string, any>} user_post The user state object, after the hunt.
+     * @param {Object <string, any>} hunt The journal entry corresponding to the active hunt.
+     */
+    function addValourRiftStage(message, user, user_post, hunt) {
```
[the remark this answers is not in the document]
groupsky: Fixed

src/scripts/main.js
```diff
@@ -1735,6 +1761,36 @@
...
+        const attrs = user.environment_atts;
+        message.hunt_details = {
+            // Elexir Rain
+            er: attrs.active_augmentations.er,
```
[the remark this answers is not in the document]
groupsky: Updated buff names

src/scripts/main.js
groupsky: I meant 1-7. There is an `is_at_eclipse` flag, so it is possible when at eclipse the floor not to be increased. For safety I added the flag to details.

groupsky: Checked the state for eclipse, it's `floor_type=8`, so no need to record `is_at_eclipse`

src/scripts/main.js
```diff
@@ -1735,6 +1766,38 @@
...
+     * @param {Object <string, any>} message The message to be sent.
+     * @param {Object <string, any>} user The user state object, when the hunt was invoked (pre-hunt).
+     * @param {Object <string, any>} user_post The user state object, after the hunt.
+     * @param {Object <string, any>} hunt The journal entry corresponding to the active hunt.
+     */
+    function addValourRiftHuntDetails(message, user, user_post, hunt) {
+        const attrs = user.environment_atts;
+        message.hunt_details = {
+            // Elexir Rain
+            elexir_rain: attrs.active_augmentations.er,
```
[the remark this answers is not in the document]
groupsky: Fixed the typo

groupsky:
@tehhowch Currently each floor is recorded as a separate stage. With expected runs going to 80+ floors, that will produce 160+ different stages. I think that will introduce issues with some of the visual tools. AFAIK the population differs only on the first 15 floors, after that it is repeated, but I don't have strong evidence for it. Should I update to track 16+ floors as one in 8-15 range?

groupsky:
Update: disabled recording of hunt details when outside

DevJackSmith:
some good suggestions on discord in #community-tools for stages part. I'm ok with hunt_details.floor = floor# for now so we can gather more info if you want before getting to stages.

Specific buckets suggested for stage are:
* Outside Tower
* Eclipse Floors
* Floors 1-7
* Floors 9-15
* Floors 17-23
* Floors 25+
And same stages for UU runs?

groupsky:
@DevJackSmith I have no observations, if not too much work for you I suggest to start with each floor separate, and when we have enough data to decide, to merge. The eclipse floor can be done now.

groupsky:
@tehhowch @DevJackSmith updated floors, now we have:
* Outside
* Floor Eclipse
* Floor 1
...
* Floor 25+

and the same with `UU` in front.

Added the floor to details, and if still in doubt I suggest to merge only the details gatherer, so we can start collecting actual data.

src/scripts/main.js
```diff
@@ -1396,6 +1397,45 @@
...
+     * @param {Object <string, any>} user The user state object, when the hunt was invoked (pre-hunt).
+     * @param {Object <string, any>} user_post The user state object, after the hunt.
+     * @param {Object <string, any>} hunt The journal entry corresponding to the active hunt.
+     */
+    function addValourRiftStage(message, user, user_post, hunt) {
+        const attrs = user.environment_atts;
+        switch (attrs.state) {
+            case 'tower':
+                var floor, prefix;
+                if (attrs.floor % 8) {
```
tsitu: This needs an `=== 0`

src/scripts/main.js
```diff
@@ -1396,6 +1397,45 @@
...
 
+    /**
+     * Report the tower stage - outside, eclipse, tower floor 1-7, 9-15, 25+-31+, umbra run
+     * @param {Object <string, any>} message The message to be sent.
+     * @param {Object <string, any>} user The user state object, when the hunt was invoked (pre-hunt).
+     * @param {Object <string, any>} user_post The user state object, after the hunt.
+     * @param {Object <string, any>} hunt The journal entry corresponding to the active hunt.
+     */
+    function addValourRiftStage(message, user, user_post, hunt) {
+        const attrs = user.environment_atts;
```
tsitu: Not sure if this affects anything, but we typically use `user.quests` to parse client-side state. `user.environment_atts` is probably updated just as frequently, but for consistency's sake it might be good to stick with `user.quests.QuestRiftValour`

src/scripts/main.js
```diff
@@ -1396,6 +1397,45 @@
...
+                    floor = 'Eclipse';
+                } else if (attrs.floor < 25) {
+                    floor = attrs.floor
+                } else {
+                    // Convert 25+ into 25-31
+                    floor = (((attrs.floor - 25) % 8) + 25) + '+'
+                }
+
+                if (attrs.active_augmentations.tu) {
+                    prefix = "UU Floor";
```
tsitu: I personally prefer leaving this as `"Umbra Floor"`

src/scripts/main.js
tsitu: Hmm, it doesn't look like `user.quests` has `active_augmentations` (at least in my case - I'm currently inside tower without any active). Might be worth making an exception then.

src/scripts/main.js
```diff
@@ -1396,6 +1397,45 @@
...
+                } else {
+                    prefix = "Floor";
+                }
+
+                message.stage = prefix + ' ' + floor
```
tsitu: I (and several other Discord users) think a plain `Eclipse` would be better than `Floor Eclipse`

src/scripts/main.js
```diff
@@ -1736,6 +1777,34 @@
...
+     * @param {Object <string, any>} user The user state object, when the hunt was invoked (pre-hunt).
+     * @param {Object <string, any>} user_post The user state object, after the hunt.
+     * @param {Object <string, any>} hunt The journal entry corresponding to the active hunt.
+     */
+    function addValourRiftHuntDetails(message, user, user_post, hunt) {
+        const attrs = user.environment_atts;
+        // outside tower the active_augmentations is not defined
+        if (attrs.stage === 'tower') {
+            message.hunt_details = {
+                elixir_rain: attrs.active_augmentations.er,
```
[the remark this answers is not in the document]
tsitu: Actually, I don't think we need to track all 6 augmentations.

Super Siphon and String Stepping might have a bit of use for hardcore analysis, but the other 4 we can definitely do without.

- Umbra: Already tracked in stage
- Elixir Rain: We know it adds 50% Elixir droprate to every mouse
- Sigil & Secret: We know they add 50% more resources to floor cache

tsitu:
@groupsky What are your thoughts on my changes?

I decided to group the floors together based on feedback from Discord. With the previous method (separate stages for each floor) we would have 29 * 2 = 58 total stages to contend with. This would be OK for floor-specific mice like Champions, but generic mice like Terrified Adventurer would return dozens of similar results with low sample sizes.

src/scripts/main.js
groupsky: yes, the quests doesn't have the active_augmentation. It does have the other attributes, but the game code uses the env_attrs, so I think it's safer to use that.

src/scripts/main.js
```diff
@@ -1736,6 +1780,25 @@
...
+     * @param {Object <string, any>} user_post The user state object, after the hunt.
+     * @param {Object <string, any>} hunt The journal entry corresponding to the active hunt.
+     */
+    function addValourRiftHuntDetails(message, user, user_post, hunt) {
+        const attrs = user.environment_atts;
+        // active_augmentations is undefined outside of the tower
+        if (attrs.stage === "tower") {
+            message.hunt_details = {
+                super_siphon: attrs.active_augmentations.ss, // 'true' when active, 'undefined' when inactive
+                string_stepping: attrs.active_augmentations.sste, // ^
```
groupsky: Not sure about the backend code - maybe the inactive state should be omitted or converted to `false`

groupsky:
@tsitu looks good

src/scripts/main.js
groupsky: `false` is more appropriate
```suggestion
                super_siphon: !!attrs.active_augmentations.ss, // 'true' when active, 'undefined' when inactive
                string_stepping: !!attrs.active_augmentations.sste, // ^
```

tsitu: Good catch

PR #70, opened 2019-10-24
Check for some possible flags set when user has KR

tehhowch:
I'd like some extra eyes on this -- thoughts welcomed, too.

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -219,7 +219,17 @@
                     },
                     dataType: "json"
                 }).done(userRqResponse => {
-                    if (debug_logging) {window.console.log({message: "Got user object, invoking huntSend", userRqResponse});}
+                    // The user may have received a King's Reward.
+                    if (userRqResponse.success !== 1 || userRqResponse.puzzle === true) {
+                        // Avoid sending an additional request (which will also fail).
+                        if (debug_logging) {
+                            window.console.log("Unable to sound the horn, perhaps there is a King's Reward?");
+                        }
+                        jqx.abort(); // Needed? Or can we just `return` and have the UI update with the KR message?
+                        return;
+                    } else if (debug_logging) {
+                        window.console.log({message: "Got user object, invoking huntSend", userRqResponse});
+                    }
                     hunt_xhr.addEventListener("loadend", () => {
                         if (debug_logging) {window.console.timeEnd("Overall 'Hunt Requested' Timing");}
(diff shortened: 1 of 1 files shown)


DevJackSmith:
what's the plan here if we do find a flag? I don't want to dive into this as other, not-so-great peeps may then copy us to improve their shady KR scripts. (That's my only concern here). I feel like MH already has a problem with this area, and i don't want to make it worse.

tsitu:
Judging by tehhowch's comments, it looks like this check is intended to prevent an additional request from being sent out, which we know will fail if the user has a KR.

I'm not 100% sure what request is being referred to though - is it the one to Jack's server?

tehhowch:
We query the user object before the hunt by catching the hunt request before it gets sent and delaying it until we receive the user object back.

This PR adds a conditional check before invoking the original hunt request, and if the hunt request _cannot_ be successful, avoids sending it.

As far as "we shouldn't do this because then bad actors might figure out how to do it too," I don't see that as a valid excuse--it's a trivial check to derive. Also, the extension *already* lets the user know when they have a KR--this is just a change that will let the extension know a little earlier in the process than it currently does.

tehhowch:
I haven't tested the behavior in this PR, so testing feedback would be helpful before we merge it (hence my code comment)

DevJackSmith:
it's all a sum of trivial derivations. Just like any lock can be picked, and safe can be cracked, etc. It's just a matter of how much motivation the user has. I'm ok with this, just saying we shouldn't take down any extra obstacles (however small they might seem) for the not-so-good intention-ed people. (just my opinion for the future of this project, i think you guys are on the same page)

src/scripts/main.js
```diff
@@ -219,7 +219,17 @@
                     },
                     dataType: "json"
                 }).done(userRqResponse => {
-                    if (debug_logging) {window.console.log({message: "Got user object, invoking huntSend", userRqResponse});}
+                    // The user may have received a King's Reward.
+                    if (userRqResponse.success !== 1 || userRqResponse.puzzle === true) {
```
tehhowch: This check of `success` would prevent anyone with an expired session (or for whom we don't correctly forward session auth) from submitting a hunt request (see https://github.com/mh-community-tools/mh-helper-extension/issues/74), so probably we should only use the `puzzle` check.

tehhowch:
Closing this as the added code path doesn't appear to get hit anyway

PR #72, opened 2019-10-24
Add basic Halloween hunt details

tehhowch:
Enables DB queries to distinguish between hunts with and without firing cannons

example hunt detail object for my horn from Gnawnia Rift with Lockbox Limburger armed:
image

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1471,6 +1471,7 @@
         // TODO: Apply any global hunt details (such as from ongoing events, auras, etc).
         [
             addEggHuntDetails,
+            addHalloweenHuntDetails,
             addLNYHuntDetails,
             addLuckyCatchHuntDetails
         ].forEach(details_func => details_func(message, user, user_post, hunt));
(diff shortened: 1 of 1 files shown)


tsitu:
What are the benefits of tracking those 3 in hunt details?

tehhowch:
AFAIK, you can't determine if the GGG should be dropped since it isn't dropped when the cannon is active.

tsitu:
Oops, had a couple of comments on this that I forgot to share.

First, is getting a precise GGG drop rate worth the effort? I hear the devs are using the global drop mechanic again, which means DR is normalized based on AR and CR (similar to Chrome Bits). This makes things quite a bit trickier to derive.

Second, even if we decide to track GGG drop rate, are the `halloween_hunt` and `in_stockpile` details strictly necessary? The former can be queried for using event start/end timestamps, and the latter we know from in-game information.

tsitu:
@tehhowch

tehhowch:
As far as the utility of those flags, they may make the actual db queries simpler (one doesn't need to remember which locations were stockpiles in which year, nor do the event timestamps need to be included in the query).

One thing that we don't yet know is if the Halloween quest will disappear immediately when the "active" portion of the event ends (probably not), or if it will stick around until the user's spooky aura is consumed, or {{insert other condition}}. 

If the details aren't useful, we can revert the PR

PR #76, opened 2020-01-10
Submit Desert Heater Base loot procs as convertible

tsitu:
With the influx of Discord screenshots showcasing Desert Heater Base's ability to generate jackpot loots such as FW trap/portal parts, Launch Boxes, H/I essences, Aromatic Oils and more, I thought it'd be useful to piggyback off of the existing active horn response handling to parse DHB's bonus journal entry and submit it as a standard convertible.

[Sample HAR file for testing](https://github.com/mh-community-tools/mh-helper-extension/files/4044903/desert_heater_base.zip)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -475,14 +475,31 @@
         if (!message.isNew || !message.messageData || !message.messageData.items || message.messageData.items.length === 0) {
             return;
         }
-        const items = message.messageData.items;
 
+        const items = message.messageData.items;
+        submitConvertible(convertible, items, response.user.user_id);
+    }
+
+    /**
+     * @typedef {Object} HgItem
+     * @property {number} item_id HitGrab's ID for this item
+     * @property {string} name HitGrab's display name for this item
+     * @property {number} quantity the number of this item received or opened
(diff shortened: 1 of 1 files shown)


tehhowch:
Should work OK. I opened #77 to update this branch to be more explicit / guarded against blindly splitting and indexing the result, in case the format changes. Currently, neither this PR or #77 will handle more than a single kind of item being dropped -- not sure if that's gonna be a thing.

@DevJackSmith what sorts of journal entries are you not interested in tracking? There are a number of bonus loot journals across different setup combinations, but not all are equally interesting (such as ones where the outcome is guaranteed, like a +1 mist).

tsitu:
Oops, didn't notice that you opened your PR from this branch. Will merge this in now and take the conversation about additional bonus journal entries to https://github.com/mh-community-tools/mh-helper-extension/issues/78.

Re: handling more than a single item from DHB - we haven't seen an instance of this so far, and I highly doubt we will. I can't think of another piece of equipment that generates multiple items either. Without an example to test on I don't think it's necessary/possible to properly handle that.

PR #79, opened 2020-01-15
fix: avoid TypeError when determining LNY state

tehhowch:
the shape of LNY.mice changed to an array, so accessing mice.costumed_* is now invalid

I have an example JSON blob of the "seeking costumed mice in order" state here: https://github.com/tehhowch/mh-helper-extension/commit/ad60f77b5ea559448711bc5fa1e21c17b5c021e5 (i removed some of the "rows" values)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -738,8 +738,7 @@
         }
 
         const quest = getActiveLNYQuest(user.quests);
-        // TODO: check if the last costumed mouse is caught, not the specific one.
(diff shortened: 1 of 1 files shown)


tehhowch:
I checked the other properties that we reference and they checked out OK, so this should be the only issue needing fixing in order to track LNY hunts & custom luck

tsitu:
LGTM. Could have also checked for `"is_caught": null` on Rat

tehhowch:
This should also work for next year, provided mice remains an array 🤞

tsitu:
Good point! 🤞

PR #81, opened 2020-01-29
Fix breaking changes introduced by the 2020.01.28 patch

tsitu:
Renamed relevant instances of `user.location` to `user.environment_name` in order to fix the current invalid `message.location`

If you notice any other breaking changes introduced with today's patch, feel free to commit on top of this PR.


tsitu:
This should have been the only change, according to [cardboardshark](https://discordapp.com/channels/275500976662773761/355474934601875457/671896529790828585). Should be good to go then.

PR #88, opened 2020-09-09
Add stage and hunt details for floating islands

Mistborn94:
This adds stage info and modifier info for Floating Islands. I have only tested this on the Law island, so more testing is probably required. 

Stage is the island display name eg. "Law island"

Hunt details contains the following:
*  Island Modifiers. These are the system identifiers for active modifiers eg. [ "empty_sky", "sky_cheese", "ore_bonus" ]. I could investigate changing this to the display name if needed. 
* BeforeWarden, AtWarden, AfterWarden - booleans
* enemy - The warden type if present

Any feedback welcome :)


Mistborn94:
Final message looks like this: 
image

Mistborn94:
After discussion with AardWolf , the modifiers has been flattened:
image

PR #89, opened 2020-09-26
Made Fragments of the Eclipse singular

DevJackSmith:


DevJackSmith:
needs db clean up to move all plural to single by id, after extension is released and previous version removed from allowed config versions on server.

tehhowch:
@DevJackSmith was the db cleanup completed?

PR #103, opened 2020-10-18
Pillage tracking

AardWolf:
Adds pillage information to hunt_details


AardWolf:
Also adds some protection against hunt_details being overwritten. Removes warden status tracking from floating islands.

Issue #92, opened 2020-09-29
Track Pillages


AardWolf:
An entry can be made in hunt_details in the case of a pillage with one of three types:

pillage_gold
pillage_points
pillage_cheese

They would have the value of the number of the thing pillaged.

AardWolf:
```
{
  "render_data": {
    "image": [
      
    ],
    "entry_id": 112835,
    "mouse_type": "sky_dancer",
    "css_class": "entry short passive catchfailuredamage",
    "entry_date": "3:15 pm",
    "environment": "Floating Islands",
    "social_link_data": {
      "slt": "p",
      "sls": 1111111,
      "slu": 2222222222,
      "sltu": "stream",
      "slst1": "journal_PassiveCatchFailureWithDamage",
      "slst2": null,
      "slst3": null
    },
    "entry_timestamp": 1601410500,
    "text": "I returned to check my trap, but it appeared A <a href=\"https://www.mousehuntgame.com/\" onclick=\"javascript:hg.views.MouseView.show('sky_dancer'); return false;\">Sky Dancer Mouse</a> ate a piece of cheese without setting off my trap.<br /><br />Additionally, the fiend pillaged 1,222 gold from me!"
  },
  "publish_data": {
    "type": "journal",
    "params": {
      "user_id": 666666,
      "journal_id": 112835
    },
    "quote": null,
    "params_string": "user_id=666666&journal_id=112835"
  },
  "wall_actions": {
    "text": "Join the Hunt",
    "href": "https://www.mousehuntgame.com/hunterprofile.php?snuid=1111111&slt=p&sls=1111111&slu=2222222222&sltu=stream&slst1=journal_PassiveCatchFailureWithDamage"
  }
}
```

Gold pillage - this is one of the journal entries (personal info masked in an obvious way)

AardWolf:
Logic could be added [here or in a function called from here](https://github.com/mh-community-tools/mh-helper-extension/blob/946faebaed3c0f4e5b58bd789282c4abb312d07d/src/scripts/main.js#L770) to parse the `journal.render_data.text` for something like `pillaged ([0-9,]+) (gold|points|pieces of bait)` (need to confirm the bait pillage message). If there's a match can then set hunt details:

```
message.hunt_details = { 
  'pillage_type': match[1],
  'pillage_amount': match[0]
}
```

AardWolf:
It would seem a bait pillage has a different message:

[...]

PR #111, opened 2020-11-11
refactor(hunt_details): always perform hunt_details calculations via `addHuntDetails` method

tehhowch:
- move pillage tracking from `createMessageFromHunt` into own method, like lucky catches
 - revert usage of `Object.assign` in location-specific hunt details methods (guaranteed no existing hunt details)
   - fixes bug in Harbour hunt details

Ref #92 #103

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -706,7 +706,6 @@
     function createMessageFromHunt(journal, user, user_post) {
         const message = {
             extension_version: formatVersion(mhhh_version),
-            hunt_details: {}
         };
         const debug_logs = [];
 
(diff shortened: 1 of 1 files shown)


AardWolf:
I don't understand why you moved away from Object.assign - that was added because otherwise the hunt details get overwritten by things - like in claw shot on a poster if the boss pillages the pillage information or poster information would be submitted but not both.

tehhowch:
@AardWolf the flow was previously that a message may not have hunt details. If details were available for a location, they would be created and then any global details would be added.

In the pillage tracking PR, every message was given hunt details by default, so the location-specific details needed to be changed to append to the existing details object, rather than create them.

With this PR, we return to the flow of the former, where not all messages will have hunt details - only those where a location trigger fired, and/or where a global condition was met, such as catching a mouse or being pillaged.

tehhowch:
@AardWolf I updated the logic to make it more clear. We now only set `hunt_details` in a single location, and explicitly merge all details objects that might be present.

AardWolf:
Thank you! It all makes sense to me and looks like it's ready to go.

AardWolf (approved):

Issue #109, opened 2020-11-03
Only strip component names at the end - Trap, Charm, Cheese, Base


AardWolf:
https://github.com/mh-community-tools/mh-helper-extension/blob/0e4a9c81b159dcbfefbbb71922cfe22d23a5322a/src/scripts/main.js#L738

This should be `/ cheese$/i`. The others could use a similar fix.

PR #112, opened 2020-11-11
fix: Avoid stripping trap/base/charm/cheese unless at end of string

tehhowch:
Ref #109 

Things still parse properly:
image


tsitu:
Simple, LGTM. Closes #109

PR #114, opened 2020-11-13
feat: create a development environment

tehhowch:
This PR doesn't do too much for us just yet, but is step 1 in a longer process that will introduce tests and better code quality

This PR
1. allows us to declare development tooling, via `devDependencies`
2. Adds a small bit of code style enforcement
   - I chose not to enable rules for quote type or quoted object literal props, as we use both styles, and doing so here is a lot of noise
3. Creates two github workflows that provide some token level of feedback to contributions
   - the lint job will fail if code is not valid javascript
   - the test job will be useless until we have tests, but when we do (and make the package.json test script run them, then GitHub will automatically run tests on push & PR
4. Create a jsconfig.json project config, for those who use VS Code, which provides type signatures for the Chrome / WebExtensions APIs

In the future, we can use this development environment to introduce a bundling/compiling process for generating releases, which means we can split up `main.js` into testable components. We can possibly also explore virtual DOM testing solutions, and inject parts of the code. It may also be possible to use a sandbox browser env, to actually test the extension's bootstrapping. (I haven't looked into any kinds of integration testing solutions for browser extensions, but I can only assume they exist.)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,4 +11,16 @@ indent_style = space
 indent_size = 4
 trim_trailing_whitespace = true
 
+[*.{yml,yaml}]
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace = true
+
+[*.md]
+indent_size = 2
+indent_style = space
+trim_trailing_whitespace = false
+
 # TODO: settings for .html files, .css, etc.
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -14,6 +14,14 @@ module.exports = {
         ecmaVersion: 6,
     },
     rules: {
+        indent: [
+            'error',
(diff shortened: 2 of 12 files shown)


AardWolf:
Thank you, I appreciate this and still benefit from a similar change to the timer bot environment.

When we get around to breaking up main.js we should be able to add a testing mechanism whereby we can feed in recorded json objects - such as a pre-hunt response and a hunt response. Or a convertible opening. No more waiting for horns / trap checks. And maybe we can add eruptions without having to actually hunt through eruptions to test it.

AardWolf (approved):

DevJackSmith:
looks good to me, one question, why are there two checks with the same name? should we maybe name them differently?
image

DevJackSmith (approved):

PR #117, opened 2020-12-15
Use a saner table schema

tehhowch:
So we can more easily do the things we wish to do.

The backend & BigQuery are all set up & verified.. just needs a deployment!


tsitu:
LGTM, thanks for implementing this! 👍

PR #119, opened 2020-12-20
Added N/A stage for non-event cheese to GWH for better aggregation

AardWolf:
Did single tests - one with PP and one with ESB. Both submitted what I expected. I'm going to assume the boss stage still works.


AardWolf:
Also tested with GPP and got expected results.

tehhowch (approved):

PR #120, opened 2021-01-08
Sleigher tracking

AardWolf:
Added a hack to track those extra journal entries that aren't convertibles. Please make sure it's not an awful hack.

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -776,24 +776,24 @@
                     }
                 }
             }
-            else if (Object.keys(journal).length !== 0) {
-                // Only the first regular mouse attraction journal entry can be the active one.
-            }
-            else if (css_class.search(/linked|passive|misc/) !== -1) {
-                // Ignore any friend hunts, trap checks, or custom loot journal entries.
-            }
-            else if (css_class.search(/(catchfailure|catchsuccess|attractionfailure|stuck_snowball_catch)/) !== -1) {
-                if (css_class.includes('active')) {
-                    journal = markup;
-                }
-                else if (Object.keys(journal).length !== 0) {
-                    done_procs = true;
-                }
-            }
             else if (!done_procs && css_class.search(/pirate_sleigh_trigger/) !== -1) {
                 // SS Scoundrel Sleigh got 'im!
                 more_details['pirate_sleigh_trigger'] = true;
-                if (debug_logging) {window.console.warn({record: more_details});}
+                if (debug_logging) {window.console.log({record: more_details});}
+            }
+            else if (css_class.search(/(catchfailure|catchsuccess|attractionfailure|stuck_snowball_catch)/) !== -1) {
+                if (debug_logging) {window.console.log({message: "Got a hunt record"});}
+                if (Object.keys(journal).length !== 0) {
+                    done_procs = true;
+                    if (debug_logging) {window.console.log({done_procs: done_procs});}
+                }
+                else if (css_class.includes('active')) {
+                    journal = markup;
+                    if (debug_logging) {window.console.log({message: "Found the active hunt", journal: journal});}
+                }
+            }
+            else if (css_class.search(/linked|passive|misc/) !== -1) {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -559,6 +560,17 @@
         sendMessageToServer(db_url, message);
     }
 
+    // Add bonus journal entry stuff to the hunt_details
+    function addMoreDetails(hunt) {
```
tehhowch, line 564: This name suggests the function parameter is modified

src/scripts/main.js
```diff
@@ -559,6 +560,17 @@
...
+        const new_details = {};
+        if ('more_details' in hunt) {
+            hunt.more_details.forEach(detail => {
+                new_details[detail] = true;
```
tehhowch, line 568: This approach will work so long as all hunt_details we want to add are booleans. Not sure how robust that assumption is.

That said, this function is effectively the same as `Array#reduce` - take an array and accumulate all its values in the same way.

src/scripts/main.js
```diff
@@ -764,6 +777,11 @@
                     }
                 }
             }
+            else if (css_class.search(/pirate_sleigh_trigger/) !== -1) {
+                // SS Scoundrel Sleigh got 'im!
+                more_details.push('pirate_sleigh_trigger');
+                if (debug_logging) {window.console.warn({record: more_details});}
```
tehhowch, line 783:
```suggestion
                if (debug_logging) {window.console.log({record: more_details});}
```

tehhowch (commented):
The biggest issue with bonus journal entries has always been what happens with trap checks and the order of entries.

I think a more stringent check on the total contents of the received journal is necessary, to avoid counting a sleigher proc when it may have happened for a friend hunt or trap check, in case those journal entries are included in the active hunt response.

AardWolf:
Regarding trap checks and other passive hunts, we don't submit any hunt information if there's more than one hunt detected. I believe we do still submit the convertibles that we track.

Also, I agree about the booleans. The more details thing doesn't need to be an array - it could be key value pairs and just copied in. That'd be more flexible later.

tehhowch:
> Regarding trap checks and other passive hunts, we don't submit any hunt information if there's more than one hunt detected. I believe we do still submit the convertibles that we track.

This is incorrect. We submit only the active hunt, but others may be present in the received markup.

The issue is that we look at *all* of these received journal entries, which means we will look at "bonus" journals associated with trap and friend hunts. Thus you could end up with sleigher_proc for a friend hunt, which wouldn't get submitted, or a sleigher_proc on a submitted active FTC because it proc'd on the trap check before your hunt, etc.

AardWolf:
> This is incorrect. We submit only the active hunt, but others may be present in the received markup.

You're right - I misread the else if chain. Once there's something in journal we stop looking for hunts to replace it.

The most recent commit appears to be working such that once any other (active) hunt is encountered it stops checking for these types of bonus entries as well.

Some things I did learn:
* if you have a stale window and hunt and your horn was actually ready, all returned journal entries show as active (unless they were friend or trap check). Even hours-old hunts which you led.  - This means all those convertibles we already sent could be sent again.

* If you have a stale window and hunt and your horn was not ready you get an error message and we handle this just fine.

tehhowch:
"active" is just a CSS class name that helps tell the browser how to format the text. It's a guarantee only that the player initiated that hunt.

We can add more robust time-of-hunt checks, though I believe the array order of entries was robust in the original testing.

AardWolf:
I proposed also added the entry's timestamp for convertible submissions in #121 and I think that's all we really need to be safe about it.

AardWolf:
I have some debug logging to clean up and to update the manifest but in my current testing I'm only getting the sleigher flag submitted as a hunt detail when it's related to the most recent hunt. I'm still hoping for a proc on a trap check followed by no proc on the active hunt to confirm that particular case.

AardWolf:
@tehhowch and others - can you please give it a look before I do something crazy like release it?

PR #124, opened 2021-03-03
Sbfactory2021

AardWolf:
Pending review and testing.

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -462,7 +462,7 @@
             sandy_bert_bites_snack_pack:	130004,
             hollow_heights_party_pack_snack_pack:	130005,
             riftios_snack_pack:	130006,
-        }
+        };
         const convertible = {};
         if (purchase.type in packs) {
             convertible["name"] = purchase.type;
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -274,6 +274,12 @@
             if (Date.now() < 1611878400000) {
                 getSettings(settings => recordGWH2020Golems(settings, xhr));
             }
+        } else if (url.includes("mousehuntgame.com/managers/ajax/events/birthday_factory.php")) {
+            // Triggers on Birthday Items claim, room change click (+others, perhaps).
+            // (GMT): Mar 24 2021 15:00:00 GMT+0000 [Birthday 2021 placeholder end date])
+            if (Date.now() < 1616598000000) {
```
DevJackSmith, line 280: Just for the future, if we are using something (for example Date.now()) more than once, we should do it once and store it in a variable. (Code reuse best practices i believe)

DevJackSmith: I'm also not a fan of us hardcoding dates throughout the extenstion. It might be better to have them at the very top of the extenstion, so we can see and change them easily if needed.

Another tip is to not use one time functions. We try to make  a function so it works for years to come, not just this year. (As you probably already know).

DevJackSmith (approved):
Interesting work around for cheese ids. I'm pondering if maybe using existing json only endpoint on server for looking up cheese id is better, but this works.

src/scripts/main.js
AardWolf: I 100% agree. I followed what we did for golem submissions and in this case it's to short-circuit any late vending machine purchases that might somehow trigger. It's probably not really necessary but feels safe. And means we won't submit next year's blindly.

AardWolf:
> Interesting work around for cheese ids. I'm pondering if maybe using existing json only endpoint on server for looking up cheese id is better, but this works.

I hate this and want to change how convertible submissions works - look up by name if we don't have an id. Create a new id starting at some high number (50K?) within convertible_intake if we have something with a new name.

PR #126, opened 2021-03-08
Fixed gauntlet string cheese proc id in riftios

AardWolf:


AardWolf:
"settings" was added to debug log to fix unused variable - and for me to debug something.

PR #127, opened 2021-03-27
Escape button closes UI elements

Hazado:


tehhowch:
Tweaks and such to the MouseHunt UI are generally reserved for their own user script and installed via tampermonkey/etc.

DevJackSmith:
@tehhowch i agree with you, but lets try to be more inclusive. I'm all for new features and options. Especially since this is disabled by default.

PR #130, opened 2021-04-17
Updated pipeline to pull in Tsitus latest tools and use them locally

DevJackSmith:


DevJackSmith:
Seems there are still bugs with this one, only Menu and CRE works for me. Since I was testing on CRE, i thought it would work for all scripts, but for some reason my FF blocks others with this message:
```Security Error: Content at https://www.mousehuntgame.com/camp.php may not load or link to moz-extension://b35ddbf9-97fe-664c-b8c6-d40c8a3c13c2/third_party/tsitu/bm-analyzer.js.```
```<script> source URI is not allowed in this document: “moz-extension://b35ddbf9-97fe-664c-b8c6-d40c8a3c13c2/third_party/tsitu/bm-analyzer.js”.```

DevJackSmith:
fixed it

PR #133, opened 2021-05-28
Added logic for opening prize packs

AardWolf:
Watches the prize pack opening endpoint

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -277,6 +277,12 @@
             if (Date.now() < 1616598000000) {
                 getSettings(settings => recordSnackPack(settings, xhr));
             }
+        } else if (url.includes("mousehuntgame.com/managers/ajax/events/kings_giveaway.php")) {
+            // Triggers on Birthday Items claim, room change click (+others, perhaps).
+            // Wed Jun 23 2021 22:00:00 GMT-0400 [King's Giveaway Key Vanishing date 15th])
+            if (Date.now() < 1624500000000) {
+                getSettings(settings => recordPrizePack(settings, xhr));
+            }
         }
     });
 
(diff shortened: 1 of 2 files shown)


AardWolf:
Test results from 1 and 25 keys: https://discord.com/channels/275500976662773761/355474934601875457/847816986473267280

@DevJackSmith I lost instructions for the auto-update of version

src/scripts/main.js
```diff
@@ -277,6 +277,12 @@
             if (Date.now() < 1616598000000) {
                 getSettings(settings => recordSnackPack(settings, xhr));
             }
+        } else if (url.includes("mousehuntgame.com/managers/ajax/events/kings_giveaway.php")) {
+            // Triggers on Birthday Items claim, room change click (+others, perhaps).
+            // Wed Jun 23 2021 22:00:00 GMT-0400 [King's Giveaway Key Vanishing date 15th])
+            if (Date.now() < 1624500000000) {
```
DevJackSmith, line 283: not sure if this works correctly as this is javascript which uses hunter's computer, which could have date wrong. In fact i don't think it works in another instance when we submit RH. But in my test in console i wrote  Date.now() and it worked correctly. So maybe there's something wrong with out RH date code:
`if (rh_message.entry_timestamp > Math.round(new Date().setUTCHours(0, 0, 0, 0) / 1000)) {`

DevJackSmith (approved):

src/scripts/main.js
AardWolf: I should probably stop copying this around. It makes sense when the target data source is somewhere else (golems) but maybe not for our stuff.

plasmoidia: RH is testing the timestamp on the message versus a date in UTC. This is just loosely checking a time we know it should not be possible for submissions to be made. Since Prize keys are set to be removed June 15, a week after on June 23 should give plenty of slop.

PR #141, opened 2021-06-08
refactor(workflows): Update workflow versions, steps

tehhowch:
- Update action versions to latest
 - Limit uses of `cd` in publish workflow
 - Use env variables for repeated cURL arguments


DevJackSmith (approved):

DevJackSmith:
thank you! :) i agree we should move it to npm build, good points

PR #157, opened 2021-12-07
themes aren't festive now

AardWolf:
Themes were festive, now they aren't. Updated the comparison appropriately.

Watch for "Toys yule log stat item" (for example) stage names until this propogates.


plasmoidia (approved):
LGTM

DevJackSmith (approved):

PR #158, opened 2022-01-23
Added sending items from scavenger maps

DevJackSmith:
sending scavenger map items to server, as well as map type

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -168,6 +168,15 @@
         return Object.values(mice);
     }
 
+    function getMapItems(data) {
+        const items = {};
+        $.each(data.treasure_map.goals.item, (key, item) => {
+            items[item.unique_id] = item.name;
+        });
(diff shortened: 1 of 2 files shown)


DevJackSmith:
decided to remake this to another simpler PR

src/scripts/main.js
```diff
@@ -168,6 +168,15 @@
         return Object.values(mice);
     }
 
+    function getMapItems(data) {
+        const items = {};
+        $.each(data.treasure_map.goals.item, (key, item) => {
+            items[item.unique_id] = item.name;
+        });
+
+        return Object.values(items);
```
tehhowch, line 177: The only need for the dict approach was to support removing caught mice by the id--if that's not needed then this is more to the point:
```suggestion
        return Object.values(data.treasure_map.goals.item)
          .map((item) => item.name);
```

PR #159, opened 2022-01-23
Allowing scavenger map items intake through mice map intake

DevJackSmith:
and added map type to be sent to server, just in case we want it in future.

Downside is our map_mice table will be 1.6GB instead of two 800MB tables if we keep mice and items separately.

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -153,7 +153,8 @@
     // Extract map mice from a map
     function getMapMice(data, uncaught_only) {
         const mice = {};
-        $.each(data.treasure_map.goals.mouse, (key, mouse) => {
+        const map_things = data.treasure_map.is_scavenger_hunt ? 'item' : 'mouse';
+        $.each(data.treasure_map.goals[map_things], (key, mouse) => {
             mice[mouse.unique_id] = mouse.name;
         });
 
@@ -631,6 +632,7 @@
         }
         const map = {
             mice: getMapMice(resp),
+            type: resp.treasure_map.map_type,
             id: map_id,
             name: name.replace(/ treasure/i, '')
                 .replace(/rare /i, '')


src/scripts/main.js
```diff
@@ -153,7 +153,8 @@
     // Extract map mice from a map
     function getMapMice(data, uncaught_only) {
         const mice = {};
-        $.each(data.treasure_map.goals.mouse, (key, mouse) => {
+        const map_things = data.treasure_map.is_scavenger_hunt ? 'item' : 'mouse';
+        $.each(data.treasure_map.goals[map_things], (key, mouse) => {
```
tehhowch, line 157: We're just doing a reduce here:
```js
const key = data.treasure_map.is_scavenger_hunt ? 'item' : 'mouse';
const contents = Object.values(data.treasure_map[key])
  .reduce((dict, val) => {
    dict[val.unique_id] = val.name;
    return dict;
  }, {});
```
Don't forget that the "uncaught only" logic needs to use this key, too.

tehhowch (changes requested):

src/scripts/main.js
DevJackSmith: if it's the same logic, I'd like to keep my version as it seems more beginner friendly. Or is there something i'm missing?

tehhowch: The important callout is that the `uncaught_only == true` branch appears to assume the treasure map has mice data and seems like it will not work for scavenger maps (`hunter.completed_goal_ids.mouse`).

I prefer using JS builtins to jQuery whenever possible, so I provided a native alternative. I was also abstracting the variable names, since the dictionary may now contain items and not always mice.

DevJackSmith:
i think this pr can be closed since it's so old and there were so many changes since then. But it can be used as an example to fix scavenger maps intake.

PR #161, opened 2022-02-01
fix(hunt-logic): Do not drop LNY hunts with event cheese before all LNY mice are caught

tehhowch:
That's no longer a thing!
FWIW, all of the LNY mice are still reported there, but they are now only listed with an `is_available` property (and all have the value `true`). We didn't use that property previously because the next-in-line mouse would be available even if it was the next required catch.


AardWolf:
Simple enough! Let's go to production.

AardWolf (approved):

PR #162, opened 2022-02-01
chore(deps): update node & ESLint versions

tehhowch:
- Update workflows to use the latest Node LTS, 16
 - Update the ESLint version to latest minor

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14.x]
+        node-version: [16.x]
 
     steps:
     - uses: actions/checkout@v2
@@ -49,7 +49,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14.x]
+        node-version: [16.x]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


tehhowch:
@AardWolf @DevJackSmith bump

AardWolf:
Sorry gang. I'm now aware of this but how do we plan to move it forward? Merge, pull on server, upgrade necessaries, restart? I am not sure where node is being used in the extension - feels like a server thing?

tehhowch:
> Sorry gang. I'm now aware of this but how do we plan to move it forward? Merge, pull on server, upgrade necessaries, restart? I am not sure where node is being used in the extension - feels like a server thing?

NodeJS is basically the runtime environment that all JavaScript in the GitHub workflow will execute in. The extension itself doesn't use NodeJS, it runs in the browser's JavaScript environment (e.g. V8, JavaScriptCore, Chakra, SpiderMonkey).

So there's no server changes needed IMO.

DevJackSmith:
just noticed this. thanks for the pr. This indeed runs only on github servers, since this is a github action.

DevJackSmith:
the package on the other hand is on the server. And i think it's something like npm install that installs correct versions of things

DevJackSmith:
I think we do have node on the server, but it's not getting upgraded here.

PR #167, opened 2022-04-07
Fix addBalacksCoveStage to look at quest for info

plasmoidia:
Apparently Balack's Cove was refactored to use a quest instead of the old viewing_atts for game state. Simply pull the info from there instead.


AardWolf:
The changes are minor, the lint failure is related to permissions in the repo.

AardWolf (approved):
Properly accoutns for a change in the user object

PR #172, opened 2022-06-18
feat(lint): upgrade ESLint to 8.18

tehhowch:
Use new rule added in 8.14 to detect constant conditions when working with ??, ||, and &&
 - rule will likely be added to eslint:recommended in ESLint v9

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -28,6 +28,7 @@ module.exports = {
                 outerIIFEBody: 'off',
             },
         ],
+        'no-constant-binary-expression': ['error'],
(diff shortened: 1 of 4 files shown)


DevJackSmith (approved):

DevJackSmith:
sounds good. are we passing it now? if not we should probably make changes to pass too

tehhowch:
> sounds good. are we passing it now? if not we should probably make changes to pass too

Yes, otherwise the CI checks would have failed. I also verified it would work, by adding expressions that should trigger it and running the linter locally.

PR #173, opened 2022-06-21
Fixing FG stage

DevJackSmith:
all mice currently show FG Closed. Don't know when this last time worked, so might want to wipe fg in db after we get some good data.


DevJackSmith:
i would also maybe argue that the percentage to discard shouldn't be >= 99? not sure how that's helpful, since it's not 100% so it's not transitioning yet, so why discard?

PR #181, opened 2022-09-01
Refresh and update popup styling

bradp:
Updates the layout and styling of the extension a bit. Removes dependency on Bootstrap, and makes it all fit a bit nicer.

**Before** Have to scroll, things are squashed, overall 👎 
Screen Shot 2022-09-01 at 6 29 53 PM
Screen Shot 2022-09-01 at 6 31 28 PM

**After** Fits better, is prettier, overall 👍 
Screen Shot 2022-09-01 at 6 30 02 PM
Screen Shot 2022-09-01 at 6 32 10 PM


src/css/popup.css
```diff
@@ -1,3 +1,111 @@
+* {
```
DevJackSmith, line 1: this is trading minimized bootstrap for 130 (plus more in future) lines of custom code we have to maintain. Unfortunately it doesn't add much that i can see as far as styling. Let's not do this please.

src/popup.html
```diff
@@ -5,49 +5,41 @@
...
+    <main>
         <h3><a href="https://www.mhct.win" target="_blank" class="popup-link-out">MH Community Tools</a></h3>
-        <span id="version" class="text-muted" style="font-size:small;"></span>
         <table class="table text-center table-responsive table-hover table-bordered">
         <tbody>
             <tr><td id="userhistory" class="popup-link">
                 Your Hunt History
             </td></tr>
             <tr><td id="mhmh" class="popup-link">
-                MHCT Map Helper (auto)
```
DevJackSmith, line 25: why are you removing Auto from labels? people should be able to see which ones are automatic which ones are manual input

src/popup.html
```diff
@@ -5,49 +5,41 @@
...
+                Map Helper
             </td></tr>
             <tr><td>
-                <a href="https://www.mhct.win/attractions.php" target="_blank" class="popup-link-out">Mouse Attraction Rates</a>
+                <a href="https://www.mhct.win/attractions.php" target="_blank" class="popup-link-out">Attraction Rates</a>
             </td></tr>
             <tr><td>
-                <a href="https://www.mhct.win/loot.php" target="_blank" class="popup-link-out">Looter (Drop Rates)</a>
```
DevJackSmith, line 31: why are we moving it down? Looter/drop rates is one of the most used tools if not the most used.

src/popup.html
```diff
@@ -5,49 +5,41 @@
...
-                <a href="https://www.mhct.win/reverse-mapper.php" target="_blank" class="popup-link-out">Reverse Mapper</a>
+                <a href="https://www.mhct.win/loot.php" target="_blank" class="popup-link-out">Looter <em>(Drop Rates)</em></a>
             </td></tr>
             <tr><td>
                 <a href="https://www.mhct.win/converter.php" target="_blank" class="popup-link-out">Convertibles</a>
             </td></tr>
             <tr><td>
                 <a href="https://www.mhct.win/rh-tracker.php" target="_blank" class="popup-link-out">Relic Hunter Tracker</a>
             </td></tr>
-            <tr><td>
```
DevJackSmith, line 45: you are moving these to the second page of popup to avoid scrolling, i get that, but maybe there's a better way? like instead of mapper and mapper reverse, put them both on one line and make the reverse an icon or something

src/popup.html
```diff
@@ -58,12 +50,13 @@ <h3><a href="https://www.mhct.win" target="_blank" class="popup-link-out">MH Com
                 Options
             </td></tr>
             <tr><td>
-                <a href="popup2.html" class="popup-link-out"><div class="pull-right">&gt;&gt;</div>Other Links</a>
```
DevJackSmith, line 61: isn't html encoded is usually more compatible than using special characters directly?

src/popup2.html
```diff
@@ -43,14 +50,16 @@
                 <a href="https://discord.gg/E4utmBD" target="_blank" class="popup-link-out">Discord Server</a>
             </td></tr>
             <tr><td>
-                <a href="https://mhwiki.hitgrab.com/wiki/index.php/Main_Page" target="_blank" class="popup-link-out">Wiki</a>
```
DevJackSmith, line 46: why are we switching to http??

src/scripts/content.js
```diff
@@ -1,140 +1,143 @@
...
+        return;
+    }
+    // Pass version # from manifest to injected script
+    const extension_version = document.createElement("input");
+    extension_version.setAttribute("id", "mhhh_version");
+    extension_version.setAttribute("type", "hidden");
+    extension_version.setAttribute("value", chrome.runtime.getManifest().version);
+    document.body.appendChild(extension_version);
 
-// Add flash message div
```
DevJackSmith, line 12: why are we moving this down?

src/scripts/popup.js
```diff
@@ -74,15 +74,20 @@ document.addEventListener('DOMContentLoaded', () => {
  * @param {HTMLElement} [huntTimerField] The div element corresponding to the horn countdown timer.
  */
 function updateHuntTimerField(tab, huntTimerField) {
+    const hornImage = '<img src="images/horn.png" class="horn-img">';
```
DevJackSmith, line 77: this loads horn image into memory every second, vs before we only loaded it when the response was ready. this has performance impact, no?

src/scripts/popup.js
```diff
@@ -74,15 +74,20 @@ document.addEventListener('DOMContentLoaded', () => {
...
         if (huntTimerField) {
-            if (response === "Ready!") {
-                huntTimerField.innerHTML = '<img src="images/horn.png" class="horn">';
+            if (response === "Ready!" ) {
+                if ( hornImage !== document.getElementById("huntTimer").innerHTML ) {
+                    huntTimerField.innerHTML = hornImage;
+                }
             } else {
-                huntTimerField.textContent = response;
+                huntTimerField.textContent = `Horn ready in ${response}`
```
DevJackSmith, line 90: this is not always seconds. this is sometimes KR.

tehhowch (changes requested):
Sharing some comments I had staged as well

DevJackSmith:
@bradp overall you have good ideas, unfortunately i think some of them are not needed / might not be good for long term. I like the yellow color change for example, and making the first popup page not scroll. Also check out lint results, seems it's failing. You can see inline comments from lint on the "Files changed" tab here: https://github.com/m-h-c-t/mh-helper-extension/pull/181/files

bradp:
😆 I definitely did not mean to change links from https to http.

I'll review the comments, looks like there's some good suggestions.

And I'll split some of the changes into separate PRs so will be a bit easier to just pick what we want to merge in.

Thanks!

src/popup.html
DevJackSmith: @bradp after thinking more on this, I'm ok with moving these to bottom of second page.

src/scripts/background.js
```diff
@@ -127,17 +127,15 @@ function icon_timer_updateBadge(tab_id, settings) {
...
-                        const seconds = response_int % 100;
-                        if (seconds > 30) {
-                            ++minutes;
-                        }
-                        response = minutes + 'm';
-                    } else {
-                        response = response_int + 's';
+                } else if (response_int > 59) {
+                    let minutes = Math.floor(response_int / 100);
+                    const seconds = response_int % 100;
```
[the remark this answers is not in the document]
DevJackSmith: I'm ok with whitespace changes, we can ignore them on github Files tab in the gear icon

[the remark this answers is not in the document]
bradp: 👍 reverted

src/scripts/content.js
```diff
@@ -1,140 +1,143 @@
...
-    return;
-}
-// Pass version # from manifest to injected script
-const extension_version = document.createElement("input");
-extension_version.setAttribute("id", "mhhh_version");
-extension_version.setAttribute("type", "hidden");
-extension_version.setAttribute("value", chrome.runtime.getManifest().version);
-document.body.appendChild(extension_version);
+    if (document.body == null) {
+        return;
```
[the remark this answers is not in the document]
bradp: 👍 reverted

src/popup2.html
```diff
@@ -43,14 +50,16 @@
...
+                <a href="http://mhwiki.hitgrab.com/wiki/index.php/Main_Page" target="_blank" class="popup-link-out">Wiki</a>
```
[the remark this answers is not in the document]
bradp: 👍 reverted, was an accident

src/scripts/popup.js
bradp: Updated. Good catch.

src/popup2.html
bradp: 👍 reverted, was an accident.

src/scripts/content.js
bradp: 👍 Updated

src/popup.html
bradp: 👍  Updated to re-add.

src/popup.html
bradp: 👍  Reverted

src/popup.html
bradp: 👍

src/popup.html
bradp: `>>` isn't correct for what this is though. Unicode characters have full support every where unless they're in an extended set, which isn't the case for this arrow character.

src/css/popup.css
bradp: Yes, but the 130 lines are simple and _only_ what we need. This allows us to control exactly what the CSS is rather than having to worry about an additional dependency. 

All of this is extracted directly out of the main bootstrap styles, with any changes being the styles that we're customizing.

There is minimal need for any browser-specific support here, as all our styles are supported widely.

bradp:
@DevJackSmith @tehhowch I've updated the PR with feedback & responses to some of your comments.

Feel free to un-resolve any of the feedback if you want to re-address it.

Thanks!

src/css/popup.css
DevJackSmith: there's no worry about an additional dependency. please don't touch bootstrap, there's no issue with it that we are trying to fix.

src/popup.html
```diff
@@ -5,49 +5,41 @@
...
     <link rel="stylesheet" href="css/popup.css">
     <script defer type="text/javascript" src="scripts/popup.js"></script>
 </head>
 <body class="text-center">
     <div id="horn" class="popup-link bg-warning">
-        Sound Horn: <span id="huntTimer">-:--</span>
+        <span id="huntTimer"></span>
     </div>
-    <div class="container-fluid">
+    <main>
```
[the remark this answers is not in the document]
DevJackSmith: please undo, as we are not removing bootstrap.

src/scripts/popup.js
```diff
@@ -78,11 +78,16 @@ function updateHuntTimerField(tab, huntTimerField) {
...
-                huntTimerField.innerHTML = '<img src="images/horn.png" class="horn">';
+            if (response === "Ready!" ) {
+                const hornImage = '<img src="images/horn.png" class="horn-img">';
+
+                if ( hornImage !== document.getElementById("hunt-timer").innerHTML ) {
+                    huntTimerField.innerHTML = hornImage;
+                }
             } else {
-                huntTimerField.textContent = response;
+                huntTimerField.textContent = `Horn ready in ${response}`;
```
DevJackSmith, line 90: this is not always time.

DevJackSmith:
@bradp thanks for your effort on this. While this is appreciated, unfortunately there's no reason to change most of the things that are changing, in fact keeping some things is more maintainable, upgradeable, and (debatably) easier to use. I like that you want to make this better, and I think we just need to talk about why we are doing things before doing them.

Maybe we should start over. Make an Issue ticket, we will discuss if it's needed, and if we agree that it's needed, we can then make a simple one change only PR.

DevJackSmith:
example issue/PR that i think would benefit from your work: moving links on the second page of popup to make first page not scroll

DevJackSmith:
another example: is making the yellow color more prominent
another example: changing http to https in few cases where we missed it

PR #185, opened 2022-09-27
New stale and transition branch

DevJackSmith:
* adding unobtrusive mode option to horn notification (and making it from background page instead of mh page) - credit to @haoala with their permission and mh devs permission (not supported by Firefox)
* filtering out all entries that are as old or older than pre-hunt response ones
* filtering out all hunts if trap check is in the list of new entries
* filtering out all hunts if stage changed from pre-hunt to post-hunt

Next steps:
* hunt_count is just a display for now, needs implementation on the back end and probably a logic to discard if there's two or more hunts. (In the near future, doesn't block this pr)


src/manifest.json
```diff
@@ -41,5 +41,11 @@
     "options_ui": {
         "page": "options.html",
         "chrome_style": true
+    },
+
+    "browser_specific_settings": {
+        "gecko": {
+            "id": "{801e5516-3311-4ee7-8185-7da12ffab807}"
+        }
```
DevJackSmith, line 49: on purpose

Issue #194, opened 2022-09-30
Convertibles - some items are not in inventory on open


AardWolf:
https://github.com/m-h-c-t/mh-helper-extension/blob/547538fda8fa1d039f64aa609e657cebef36076a/src/scripts/main.js#L817

Need to add an if-else here to check for gold_stat_item and point_stat_item. Map them to the appropriate item id:

gold -> 431
points -> 644

AardWolf:
Discussed on Discord: https://discord.com/channels/275500976662773761/355474934601875457/1025384862653358161

hymccord:
Tests added during #637

PR #195, opened 2022-09-30
coded fix, needed to test

AardWolf:
Tested, shared on discord. This addresses #194

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -804,6 +804,10 @@
         }
 
         const results = response.convertible_open.items;
+        const our_map = {
+            gold_stat_item: 431,
+            point_stat_item: 644,
+        };
         const items = [];
         results.forEach(result => {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -815,6 +819,15 @@
                     quantity: result.quantity,
                 });
             }
+            else if (result.type in our_map) {
+                items.push({
+                    id: our_map[result.type],
+                    type: result.type,
+                    name: result.name,
+                    pluralized_name: result.pluralized_name || '',
```
AardWolf, line 827: Should probably do this for the other pluralized name

AardWolf (commented):
Looks good to me, of course. The pluralized name getting a default is new behavior not used in the other area that sets the pluralized name. Not sure if we can count on ALWAYS getting one returned.

DevJackSmith (approved):

src/scripts/main.js
DevJackSmith: added

PR #200, opened 2022-10-01
Add typescript tooling

hymccord:
This PR adds the capability to convert JavaScript files into TypeScript and still have them in the extension.

The main change is that now the extension will need to be compiled via `npm run build` before you can load it into chrome/firefox or browser of your choice from the top level `dist/` folder. Technically you can load it from the `src/` folder now, but once typescript files are added, you must build it and load from `dist/`. 
The upside is that now you can `npm run dev` and it will launch an isolated browser profile and auto reload the extension when you make `src/` changes.

- Update README and Contributing docs for new workflow
- The publish and ci workflows will now build/compile (i.e. type check) the extension

I tested the publish workflow on my main branch here: [Publish extension](https://github.com/hymccord/mh-helper-extension/actions/runs/3165352130). This one actually scared me slightly because it says the push chrome action succeeded, however there is actually a 401 error at the end of the test (expected since secrets are not included in forks).
I tested the PR (ci) worflow with this [PR](https://github.com/hymccord/mh-helper-extension/pull/1). You can see the workflow run [here](https://github.com/hymccord/mh-helper-extension/actions/runs/3165290776).

The actual conversion to TypeScript can be done piece-by-piece in subsequent PRs. I tested this workflow out by converting `src/scripts/popup.js` in a personal PR [here](https://github.com/hymccord/mh-helper-extension/pull/3) with the resulting workflow [here](https://github.com/hymccord/mh-helper-extension/actions/runs/3165559868).


DevJackSmith:
not sure how i feel about this. What do you guys think @tehhowch @AardWolf ?

webpack/webpack.common.js
```diff
@@ -0,0 +1,54 @@
...
+                },
+            },
+        ],
+    },
+    // Setup @src path resolution for TypeScript files
+    resolve: {
+        extensions: [".ts", ".tsx", ".js"],
+        alias: {
+            "@src": path.resolve(__dirname, "../src/"),
+        },
```
tehhowch, line 37: What is this for?

webpack/webpack.common.js
```diff
@@ -0,0 +1,54 @@
+const path = require("path");
+const srcScripts = '../src/scripts/';
```
tehhowch, line 2: If you're going to use `path.join`, why are you manually joining these parts?

tsconfig.json
```diff
@@ -0,0 +1,30 @@
...
+      "module": "commonjs",
+      "moduleResolution": "node",
+      "rootDir": "./src",
+      "outDir": "dist/scripts",
+      "strict": true,
+      "sourceMap": true,
+      "jsx": "react",
+      "lib": [
+        "DOM",
+        "DOM.Iterable"
```
tehhowch, line 17: ?

tsconfig.json
```diff
@@ -0,0 +1,30 @@
...
+          "@src/*": ["src/*"]
+      },
+      "target": "es6",
+      "module": "commonjs",
+      "moduleResolution": "node",
+      "rootDir": "./src",
+      "outDir": "dist/scripts",
+      "strict": true,
+      "sourceMap": true,
+      "jsx": "react",
```
tehhowch, line 14: We should not hint that react is something we want to support in this extension
```suggestion
```

webpack/webpack.common.js
```diff
@@ -0,0 +1,54 @@
...
+        new ForkTsCheckerWebpackPlugin(),
+        new CopyPlugin({
+            patterns: [
+                {
+                    from: './',
+                    to: outpath,
+                    globOptions: {
+                        ignore: ['**/scripts'],
+                    },
+                    context: 'src/',
```
tehhowch, line 49: It is probably more useful to move asset files out of `src/` entirely, since there is no requirement for `src/` to be mountable as an extension any longer.

Then typescript won't have to even skip html / css files, and this plugin won't have to skip the script directory.

package.json
```diff
@@ -42,6 +50,19 @@
     "url": "git+https://github.com/mh-community-tools/mh-helper-extension.git"
   },
   "devDependencies": {
-    "eslint": "^8.18.0"
+    "@types/webextension-polyfill": "^0.9.1",
+    "concurrently": "^7.4.0",
+    "copy-webpack-plugin": "^11.0.0",
+    "eslint": "^8.24.0",
+    "fork-ts-checker-webpack-plugin": "^7.2.13",
+    "ts-loader": "^8.4.0",
```
tehhowch, line 58: Why not v9?

tehhowch (commented):
I'm happy to see the dev ergonomics improving. I do have some questions / suggestions

package.json
```diff
@@ -42,6 +50,19 @@
...
+    "typescript": "^4.8.4",
+    "web-ext": "^7.2.0",
+    "webpack": "^5.74.0",
+    "webpack-cli": "^4.10.0",
+    "webpack-merge": "^5.8.0"
+  },
+  "dependencies": {
+    "webextension-polyfill": "^0.10.0"
```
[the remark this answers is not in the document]
hymccord: Sorry, this was intended for the typescript portion. It's the Promised based WebExtension API.
An alternative would be to use `@types/chrome`.

An example:

```javascript
import browser from "webextension-polyfill";
```

```javascript
    chrome.tabs.query({'url': ['*://www.mousehuntgame.com/*', '*://apps.facebook.com/mousehunt/*']}, tabs => {
        // then code
    });
```

turns into

```typescript
    const tabs: browser.Tabs.Tab[] = await browser.tabs.query({'url': ['*://www.mousehuntgame.com/*', '*://apps.facebook.com/mousehunt/*']});
    // then code

File: tsconfig.json (hunk 3 above)
Comment (hymccord) [reply tsconfig.json:17]: A lot of this came from a template repo I started with. It was typescript+react. I will remove all extraneous options.

File: tsconfig.json (hunk 4 above)
Comment (hymccord) [reply tsconfig.json:14]: Same as above. Will fix.

File: webpack/webpack.common.js (hunk 5 above)
Comment (hymccord) [reply webpack/webpack.common.js:49]: I actually had this on a previous version of my branch. I had moved all of the assets to `public/` and just the scripts were in `src/`. I decided against it in case the typescript tooling didn't want to be incorporated, you could still load the extension from src without using `npm`.

File: tsconfig.json (hunk 3 above)
Comment (hymccord) [reply tsconfig.json:17]: DOM is required (I think) for DOM definitions - `window`, `document`, etc. in TS files.

File: webpack/webpack.common.js (hunk 2 above)
Comment (hymccord) [reply webpack/webpack.common.js:2]: The entries or the imports? I have change the entries to not manually join using `+`.

File: webpack/webpack.common.js (hunk 1 above)
Comment (hymccord) [reply webpack/webpack.common.js:37]: It's for easier importing of modules. But it's not used so I will remove it.

File: .eslintrc.js [hunk 8]
```diff
@@ -2,6 +2,7 @@ module.exports = {
     env: {
         browser: true,
         es2017: true,
+        node: true,
```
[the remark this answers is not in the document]
hymccord: I've removed this. I was getting eslint intellisense problems in vscode while browsing the webpack.*.js files.  
Stuff like so:
image

hymccord:
Jack left some thoughts in the discord and I agreed. The largest issue with this PR is that it _will_ require tooling to be installed for development (Node v16) if typescript are added.  

I don't want development to be a burden for those who contribute the most to the extension. However, we could have it both ways and I could remove the typescript compiling. Contributors who don't want to use `npm` can just load out of `src` and those who can/want auto-reloaded can just do `npm run dev` and it'll load it out of `dist`.

README.md
```diff
@@ -17,7 +17,23 @@ This browser extension collects crowdsourced hunt info for [MouseHunt](https://w
...
+```
+
+**Scripts**
+
+- `npm run dev` - run `build:watch` and load extension into temporary chrome profile
+- `npm run dev:firefox` - same as `npm run dev` but for firefox
+- `npm run build` - build the production-ready unpacked extension
+- `npm run build:dev` - build the debug unpacked extension
+
+First, please change the version number in `src/manifest.json` to avoid getting banned.
```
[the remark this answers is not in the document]
DevJackSmith: yes, please change this one to please keep the version number 0.0.0 to avoid getting banned.

DevJackSmith:
just to clarify, i listed some pros and cons. Since i'm not familiar with Typescript, i don't know many pros. So mostly cons:
1. can't just clone and work on it, have to run npm/etc. (this is probably not a con, since this is a proper way to dev in JS/TS)
2. if we do any kind of modifications like minimize, we need to submit not only zip for extension, but source code too.
3. adds another layer of complexity. Now if something doesn't work, i know it's in source code. After this, it might be in the npm/webpack/etc layer as well as in source code.
4. learning curve is higher for new comers, those who don't know npm/webpack/ts/etc, might not want to contribute. this is kind of same as number 1 i guess, but it's a contribution barrier and ease of use barrier, so i count this as different.
5. i donno TS, if we switch, existing maintainers will have to learn and know npm/webpack/ts. I'm ok with it, as i'm old school but don't want to stand in front of moving forward if there's a good reason.
6. More dependencies (same as other points?)

Which brings me to "good reason": I don't see one yet, I've been told there is:
A. "the main benefit is type checking. and with having types, it makes code management easier"
B. I think it also had benefit of autoreload changes when developing locally (i know that's a small one, but still)
C. Using web-ext which is specifically made for this, and will probably contribute to correctly implementing things (i'm already using a wrapper of it for submission to FF store as of few days ago)
D. More universal structure/control/organization
E. Tehhowch is not against it, which is a good sign :)

I'm ok if we want to move forward with it and complicate things a bit. I feel like it's a good enough reason to sync up to industry standard and will make our lives easier in the long run.

.eslintrc.js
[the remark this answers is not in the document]
tehhowch: That just means you need to update the ESLint config to say that _those files_ are in a node context.

README.md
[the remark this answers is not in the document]
tehhowch: Yes, the key word was _manual_. If we have a dev script that auto-launches the compiled extension, it can make it so that the version change is _automatic_.

webpack/webpack.common.js
tehhowch: (This variable is a bunch of manually-joined paths)

webpack/webpack.common.js
tehhowch: I think it makes sense to only put things in src that get compiled. I don't think it is a major concern that the dev tooling is now required to do development.

package.json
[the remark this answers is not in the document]
hymccord: I've removed it and it can be introduced later if need be.

webpack/webpack.common.js
hymccord: Got it. Fixed.

webpack/webpack.common.js
hymccord: I will leave this one to last. If it's decided the TS tooling is a go and npm is required to dev, then I can move assets to a separate file.

hymccord:
Thanks for the feedback so far 👍 
Only one or two of the regular contributors haven't said if they ability to install node. I'll abandon the TS portion if any can't, but the web-ext portion is worth keeping in any case.

I don't foresee the npm/webpack files changing much. Outside of the initial dependencies added here and adding another dependency for importing the webextension types (`chrome.runtime` etc..). Not too much needed. I haven't yet looked into linting for TS.

It's hard for me to state the benefits of a statically typed language over dynamic because almost all of my work and personal projects have always used static typing. It might be best said by others on the web.

DevJackSmith:
i don't think it's an issue of existing maintainers installing node, npm, etc. I think it's more of a bigger learning curve and larger entrance barrier for not so experienced devs to make one small pr to help out.
Overall I'm ok with it, Tehhowch seems also ok. Lets do it. Let's try your way typescript, npm, the whole lot. We can always go back if we see a huge reason for it.

.gitignore
```diff
@@ -41,3 +41,4 @@ _ReSharper*/
 /.vs
 /src/third_party/*
 mh-helper-extension.code-workspace
+dist/
```
DevJackSmith, line 44:
```suggestion
dist/

```

DevJackSmith (approved):

PR #205, opened 2022-10-03
Add .vscode recommendations for linting

hymccord:
More dev ergonomics.

Since linting will make PRs fail, this will add vscode settings to recommend the [microsoft/vscode-eslint](https://github.com/microsoft/vscode-eslint) extension and fix files on save.

The `eslintrc.js` might need to filter the paths to `src` since it resulted in the `{args: 'none'}` change in itself. This might be trivial and not worth fixing though.


hymccord:
Rebased and force pushed on latest `main`.

PR #206, opened 2022-10-05
Update extension icon if KR puzzle is active in HUD or page element

hymccord:


hymccord:
image

Found this when my window was just showing the HUD area. I was prompted with a KR but the extension icon still had a timer after clicking the 'Camp' button.

DevJackSmith (approved):
thanks!

PR #209, opened 2022-10-07
Split staging into separate module

hymccord:
First attempt at reducing the size of the main file.

- Moved all staging functions to stages.ts (added types)
- Added initial types needed in staging (User object, a sample quest object for ToC).

I'm going to leave this as a draft as it might be too big of a change to start with.
I think the whole main.js being wrapped in `(function(){ <main_code> })();` isn't necessary anymore as webpack does its own `(() => { <main.js> })();` essentially (look at the compiled main.js in dist). I left as it for now b/c it makes the diff quite big for main.


hymccord:
Actually, going to start with just adding User type first.

PR #210, opened 2022-10-07
Ges update

AardWolf:
Depot stage has a new substage. Some database cleanup can be done to move all no-rush stage hunts with the supply schedule charm armed into this new stage


AardWolf:
Addresses #211

AardWolf:
Tested - still loads and tries to submit hunts.

PR #212, opened 2022-10-10
added daredevil canyon charms

StijnJacobsPXL:
added the 3 area charms (magmatic crystal, black powder & dusty coal) to Daredevil Canyon


AardWolf (approved):
These look correct. Now discussing the stage names.

Issue #208, opened 2022-10-07
Add Handler For Pre/Post Differences


AardWolf:
A recent change calculates the submission message for both the pre and post hunt data then compares the stages. If there is a stage mismatch the hunt is not submitted. This should be expanded to include other dimensions - location and cheese most importantly.

Doing so means we would lose some well-handled examples such as catching realm ripper.

A proposal is to add a type of pass object that can be populated like:
```
passObject = { cheese: pre_message.cheese === post_message.cheese, location: pre_message.location === post_message.location, ... }
```
This would give us binary properties that we can then clear "universally" and more specifically.

A simplified universal clearer could clear cheese (and charm) if the post cheese name is missing and the pre cheese count was 1.

A specific clearer could be something like in floating islands if the hunt count on an island was 1 then it could clear many differences based on what's in the trap library for launchpad iff the post stage is launchpad.

Specific clearers could be functionally called similar to how stage calculation and hunt detail calculations are currently done.

AardWolf:
Marked high priority because one the structure and an example is there this becomes easy work for hacktoberfest.

DevJackSmith:
I don't want to rush this just because hacktoberfest

AardWolf:
but we could! I want to rush this because right now we're rejecting hunts we could be accepting.

DevJackSmith:
we need to reject more before we accept

DevJackSmith:
and we need to clean, to make sure we are not accepting anything bad

PR #216, opened 2022-10-13
Add hunt rejection engine

hymccord:
Nomenclature I used:
`HgResponse`: The JSON object received from hg servers (ie page.php or activeturn.php)
`User`: The user object returned inside every API response (response["user"])
`IntakeMessage`: The object created from pre and post responses + journal that is sent to mhct intake.php

This PR attempts to resolve differences between pre and post objects through the use of a rules engine. The effect will be that differences in message can be salvaged allowing a hunt to be submitted to the db instead of throwing out valid data.
The engine currently has functions for validating API responses, user objects, and the resulting intake messages.

API response and user rules (implementing `IRule`), are simple classes that return a boolean based on the pre and post objects passed in via `validate(pre, post)`. Most of these rules are encapsulations of simple field comparison.

Intake Message validation is the most important part of this PR since this is where most hunts currently fail. These rules implement `IPropertyRule`, which are the same as `IRule` except that the interface provides a string for which property it invalidates. This property is important since invalid IntakeMessages can be "re-validated" through the use of exemptions. The exemptions are gathered (since there may be many in the future) by what properties are currently invalid (the ones given by `IPropertyRule`.  The exemptions take the same pre and post object and return a list of properties that its own rule deems valid. Those exempted properties are removed from the invalid property list.

An example focusing on the intake messages to make it easier to understand. Hunting in Valour Rift and you're at the eclipse mouse. This will put you in the "Eclipse" stage. Hunting and catching the eclipse mouse will then put you in one of the "Floors" stage. `validateMessage` will initially fail because of the `IntakeMessageSameStage` `IFilteredRule`. This rule makes sure the stages are the same. Since they are not, the rule's property ("stage") is stored as invalid. Once all the rules have run, only the stage is invalid. The engine pulls all exemptions related to the "stage" property. There are none so the validation fails and the hunt is dumped.

Another example is Realm Ripper related. A catch of the RR in FG moves the hunter from FG to AR, the stage also changes. We have invalidations on both location and stage. Exemptions are queried and one possible match is found from the "stage" exemptions. The RealmRipper exemption asserts previous and current location and mouse caught. All match the current objects so this exemption returns all properties it expects to be invalid. In this case, it returns ["stage", "location"]. These properties are removed from the invalid list. If either one of these was left, the exemption wouldn't fully revalidate the message. Since there are no more invalid properties, we know this message is valid and we can return `true` early.

I attempted to translate (and simplify) all existing rules that were in `recordHuntWithPrehuntUser`.

This also sets up the initial typescript interfaces. See mhct.ts and hg.ts.  
Naming changes needed on any interfaces/types? Your input is welcome.

Closes: #208


src/scripts/types/mhct.ts
```diff
@@ -0,0 +1,36 @@
+export interface IntakeMessage {
+    extension_version: number;
+    user_id: number;
+    entry_id: number;
+    location: ComponentEntry | null;
```
hymccord, line 5: I really want this to not be nullable so the rules can depend on location being set. I did make it nullable _for now_ because a lot of the stage funcs will set the location to null.

Previously, the `recordHuntWithPrehuntUser` would check that location is null after the stage function so it could fail out. I think it may be better for the stage funcs to throw upon encountering an error (since it now wrapped in a try/catch) as we can assure location is _not_ null.

This null check on location is temporarily in `engine.validateMessage`.

DevJackSmith:
note to myself: i ran npm update in a separate pr, so might have to merge in main branch again, and re-add package.json changes, and do npm install/update. (to fix the conflicts)

src/scripts/hunt-filter/interfaces.ts
```diff
@@ -0,0 +1,60 @@
...
+     * Get exemptions for the given pre and post messages
+     * @param pre The pre-hunt IntakeMessage
+     * @param post The post-hunt IntakeMessage
+     * @returns An array of keys (of IntakeMessage) for which properties has been exempted or null. 
+     */
+    getExemptions(pre: IntakeMessage, post: IntakeMessage): (keyof IntakeMessage)[] | null;
+}
+
+/**
+ * Base class for all rules to gather class name for logging
```
tehhowch, line 50: Why does it need to be a base class rather than an interface? I'd prefer avoiding an inheritance hierarchy for something as trivial as "log the class name"

src/scripts/hunt-filter/engine.ts
```diff
@@ -0,0 +1,129 @@
...
+        // TODO: Refactor? Location is being set to null in stage funcs so it'll fail later
+        if (pre.location === null || post.location === null) {
+            return false;
+        }
+
+        // Run rules. Build set of currently invalid properties
+        // { "location", "stage" }
+        const invalidProperties = new Set<(keyof IntakeMessage)>();
+        for (const rule of this.messageRules) {
+            const valid: boolean = rule.isValid(pre, post);
```
tehhowch, line 58: Why the explicit type? It shouldn't be necessary IMO

src/scripts/hunt-filter/engine.ts
```diff
@@ -0,0 +1,129 @@
...
+            if (!valid) {
+                this.logger.debug(`Message possibly invalid due to rule: ${rule.name}`);
```
tehhowch, line 60: If rules implement `toString` the name doesn't need to be a public field.

src/scripts/hunt-filter/engine.ts
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        // Find exception objects that will run for given key
+        const relevant_exemptions_providers = this.messageExemptions.filter(e => invalidProperties.has(e.property));
+        this.logger.debug(`Got ${relevant_exemptions_providers.length} exemption providers for these invalid properties:`, ...invalidProperties.values());
+
+        // Do we need to filter them furthur? 
+        // Exception object could give specific key and value to filter on.
+        // Didn't pre-optimize but could be worth if there are many, many exemptions and they are slow (doubtful)
+        for (const exemption_provider of relevant_exemptions_providers) {
```
tehhowch, line 77: You can avoid the unnecessary verbosity by naming it e.g. `provider`

src/scripts/hunt-filter/engine.ts
```diff
@@ -0,0 +1,129 @@
...
+            properties: { ...invalidProperties.values() },
+            messages: { pre, post }
+        })
+        
+        return false;
+    }
+
+    private initResponseRules() {
+        this.responseRules = [
+            new ResponseRules.ApiResponseBothRequireSuccess(),
```
tehhowch, line 103: Is there a reason to require creating a rule and also remembering to create it here? Should the default export from the rule files be an array of all the rules instead?

src/scripts/hunt-filter/messageExemptions.ts
```diff
@@ -0,0 +1,31 @@
+import { IntakeMessage } from "../types/mhct";
```
tehhowch, line 1: Make sure type imports use `import type`

src/scripts/hunt-filter/engine.ts
hymccord: It will default to `Set<unknown>` if you don't specify the type. This enforces that only keys of IntakeMessage can be added to the set.

src/scripts/hunt-filter/engine.test.ts
```diff
@@ -0,0 +1,199 @@
...
+import {ApiResponse, User} from '../types/hg';
+import {IntakeMessage} from '../types/mhct';
+import {Logger} from '../util/logger';
+jest.mock('../util/logger')
+
+// Mock logger won't actually call console.log
+const logger = new Logger();
+const engine = new IntakeRejectionEngine(logger);
+
+describe('validateResponse', () => {
```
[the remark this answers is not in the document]
hymccord: I don't see it in dist. Webpack leaves it out since it is never used in any imports in the entries.

I did have it in a `./tests/scripts/hunt-filter` (ie mirror), but then you get really verbose import statements
```
import {blah} from '../../../src/scripts/hunt-filter/blah'
```
instead of 
```
import {blah} from './blah'
```

src/scripts/util/logger.ts
```diff
@@ -0,0 +1,79 @@
...
+export interface ILogger {
+    getLevel(): LogLevel;
+    setLevel(level: LogLevel): void;
+
+    trace(message: string, ...args: any[]): void;
+    debug(message: string, ...args: any[]): void;
+    info(message: string, ...args: any[]): void;
+    warn(message: string, ...args: any[]): void;
+    error(message: string, ...args: any[]): void;
+    critical(message: string, ...args: any[]): void;
```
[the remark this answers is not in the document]
hymccord: Yes, probably unnecessary since I don't think the extension can crash chrome.
I think most usage around is debug + info + error.

src/scripts/hunt-filter/interfaces.ts
hymccord: It was to cut down on the amount of lines needed to write a rule/exemption. Without a base class to set the class name, each rule needed to implement the field.

```ts
export class UserNumActiveTurnsIncrementedByOne extends RuleBase<User> {
    isValid(pre: User, post: User): boolean {
        return post.num_active_turns - pre.num_active_turns === 1;
    }
}
```
or
```ts
export class UserNumActiveTurnsIncrementedByOne implements IRule<User> {
    name: string = "UserNumActiveTurnsIncrementedByOne"
    isValid(pre: User, post: User): boolean {
        return post.num_active_turns - pre.num_active_turns === 1;
    }
}
```
or as you mentioned in another comment override `toString` 
Downside to this: even if you add `toString` to the `IRule` interface, the compiler doesn't force any classes to implement it since it's technically already implemented by Object.
```ts
export class UserNumActiveTurnsIncrementedByOne implements IRule<User> {
    isValid(pre: User, post: User): boolean {
        return post.num_active_turns - pre.num_active_turns === 1;
    }
    
    toString(): string {
        return "UserNumActiveTurnsIncrementedByOne";
    }
}
```

I'm good with swapping to option 2 (and naming it like `description`). I wanted debug logging to be able to log why rules failed and each rule provide a short description of what it does. The class name did that but it's a little ugly.
But I don't think `toString` is enforceable if a new user were to create an exemption. I want the compiler to complain, not reviewers.

src/scripts/hunt-filter/messageExemptions.ts
hymccord: Could you expound on that? I'm not sure what you mean.

src/scripts/hunt-filter/engine.ts
hymccord: Good idea. I will see what I can do.

src/scripts/hunt-filter/engine.test.ts
```diff
@@ -0,0 +1,199 @@
...
+    let pre: ApiResponse;
+    let post: ApiResponse;
+
+    beforeEach(() => {
+        pre = getValidPreResponse();
```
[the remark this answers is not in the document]
hymccord: This was again to reduce the verbosity of tests. Every test needs a new/clean pre + post object. I kept it DRY in this case.

```ts
    it('passes with good responses', () => {
        expect(engine.validateResponse(pre, post)).toBe(true);
    });
    
    it('rejects when pre is unsuccessful', () => {
        pre.success = 0;
        expect(engine.validateResponse(pre, post)).toBe(false);
    });
```

```ts
    it('passes with good responses', () => {
        let pre = getValidReponse();
        let post = getValidResponse();
        expect(engine.validateResponse(pre, post)).toBe(true);
    });
    
    it('rejects when pre is unsuccessful', () => {
        let pre = getValidReponse();
        pre.success = 0;
        let post = getValidResponse();
        expect(engine.validateResponse(pre, post)).toBe(false);
    });
```

@DevJackSmith @AardWolf What do you think?

src/scripts/hunt-filter/engine.ts
hymccord: I _can_ do this, however I think we lose the ability to inject dependencies via the constructor (maybe someone needs `ILogger` later). We can go back if needed though. Nothing lost.

hymccord:
> note to myself: i ran npm update in a separate pr, so might have to merge in main branch again, and re-add package.json changes, and do npm install/update. (to fix the conflicts)

When I merged main, I ran `npm update` as well.

src/scripts/hunt-filter/interfaces.ts
hymccord: Does anyone have a good suggestion for what the field should be named for debug logging?

src/scripts/hunt-filter/engine.test.ts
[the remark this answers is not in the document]
DevJackSmith: i'm ok with long paths, i agree they should be in separate folder. Maybe some {root_path}/src/... variable?

src/scripts/hunt-filter/engine.test.ts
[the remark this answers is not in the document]
DevJackSmith: i'm ok with beforeeach, it is a nice use of it instead of duplication in my opinion

DevJackSmith:
should we add these tests to github actions? we can do in another PR to keep this one cleaner, just need to know how to run these

src/scripts/hunt-filter/engine.test.ts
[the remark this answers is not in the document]
DevJackSmith: also, these tests are used inside the source, no? i'm confused. I thought we were doing tests to check if pre/post object matches, not tests in the sense that we test if parts of extension works. If i'm correct, then this should be part of the src, no?

the other tests like unit tests, cucumber, etc, should be in separate folder in my experience.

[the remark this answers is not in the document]
hymccord: I will move to mirror folder and create a var to shorten things up.

These tests are not part of source. They are exclusively unit testing the methods + rules.
Each test
1. Grabs a valid pre + post (ApiReponse | User | IntakeMessage)
2. Modifies a properties to make it invalid since pre and post differ
3. Runs the appropriate `validate` method.
4. Generally asserts `false`. But in the case of testing exemptions, it assert `true`.

hymccord:
Not sure where a comment went, but these tests do run as part of the pipeline already.
The previous `test` task just ran an `echo 'Nothing to see here'`.

DevJackSmith (dismissed):

src/scripts/hunt-filter/messageExemptions.ts
DevJackSmith: @tehhowch ^^

src/scripts/types/mhct.ts
DevJackSmith: in a separate pr we can fix stage checks to not use location. thanks for explaining.

src/scripts/hunt-filter/interfaces.ts
DevJackSmith: what are the field values?

hymccord: I've removed the inheritance and went with option 2. Renamed the `name` field to `description`. It's not too verbose.

@DevJackSmith The field value previously was the rule name, it was being set by the base class.
Now it's called 'description' and gives a very brief statement on what the rule/exemption does.

src/scripts/hunt-filter/messageExemptions.ts
hymccord: The only valid way here to use `import IntakeMessage from "../types/mhct"` is if IntakeMessage is a `default export` from mhct, but it is not so this is the only valid notation for importing.

src/scripts/types/mhct.ts
```diff
@@ -0,0 +1,36 @@
+export interface IntakeMessage {
```
AardWolf, line 1: Does this need a hunt_details component? We normally submit a message.hunt_details element (when there are details for the area/event)

hymccord: It will. It's not in yet because it's only done in the untyped main.js code. See `createIntakeMessage`.

AardWolf: ok, so hunt details can still submit with the message then? I saw it was still making hunt details but was worried typescript might get angry at it or something.

hymccord: Correct. If you did need to access hunt_details from typescript, then it will need to be added.
This might be done in my details PR that's in draft.

PR #222, opened 2022-10-16
updated popup styling

DevJackSmith:
- made the hunt timer bigger and bolder
- combined few buttons to make the scroll go away
- thanks for all the help @bradp 
- also moved options into gear icon
- changed hover color and to be only over cell, not whole row

image

### DIFF
--- a/src/css/popup.css
+++ b/src/css/popup.css
@@ -12,6 +12,12 @@
     color: black;
     white-space: nowrap;
 }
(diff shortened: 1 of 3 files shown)


hymccord:
All the links work. I like how the layout matches the website.

One small thing:

When you hover over a button it highlights the entire row instead of just that one button.
image

PR #233, opened 2022-10-25
Use Web Crypto API instead of forge

hymccord:
Built in API allows usage of basic hash functions without the use of third-party library

This fix was done because the main script was not being injected unless the forge library was correctly located in `dist/third_party/forge`. chrome.runtime.getURL would return undefined so the callback to load the main script never fired.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -44,14 +44,6 @@ jobs:
           curl ${{ env.curl_opts }} ${{ env.tsitu_cdn }}/bm-setup-fields.min.js
           curl ${{ env.curl_opts }} ${{ env.tsitu_cdn }}/bm-setup-items.min.js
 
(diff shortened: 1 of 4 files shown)


DevJackSmith:
thanks. (Just to clarify this was only an issue for someone who was developing extension locally, in prod version forge was included by github action.)
this is good small one purpose pr, easier to test, and review, and approve :) nicely done.

One quirk: in my tests, the code doesn't wait for `createHunterIdHash()` function to finish before continuing. Meaning that it's a race condition with other things that use user_id. (might be my tests are not accurate, i just added console log after function call, and it fired first instead of waiting, let me know if you get different results.)

DevJackSmith:
works ok in my tests

DevJackSmith (approved):

DevJackSmith:
thanks

PR #243, opened 2022-11-11
Add infra for running unit tests

hymccord:
Adding infrastructure files to allow unit testing with both `.js` and `.ts` files.

ESLint changes:
Allow exports from javascript files since browsers support it. Currently unused by any extension files.
Add the jest environemnt to js test files. Globals like `test`, `describe`, and `expect` are imported.

jest.config.js:
Jest can run without config files, but `ts-jest` is needed to compile `.ts` files.
We also configure importing path maps to simplify test file imports. i.e. `require('@scripts/file')` instead of `require('../../src/scripts/file')`)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,6 +12,7 @@ module.exports = {
     },
     parserOptions: {
         ecmaVersion: 12,
+        sourceType: 'module',
     },
     rules: {
         'array-bracket-newline': ['error', 'consistent'],
@@ -40,4 +41,15 @@ module.exports = {
         'prefer-const': ['error'],
         'semi': ['error', 'always'],
     },
+    overrides: [
+        {
+            files: [
+                '**/*.test.js',
+            ],
+            env: {
+                jest: true,
+            },
+            plugins: ['jest'],
+        },
+    ],
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -44,7 +44,7 @@ module.exports = {
     overrides: [
         {
             files: [
-                '**/*.test.js',
+                'tests/**/*.{test,spec}.js',
             ],
             env: {
                 jest: true,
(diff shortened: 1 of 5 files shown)


hymccord:
While there are no test files, I created some samples in my fork branch: https://github.com/hymccord/mh-helper-extension/tree/add-test-libraries-with-files

You can fetch and create this branch locally using your CLI:
```bash
git fetch https://github.com/hymccord/mh-helper-extension add-test-libraries-with-files:add-test-libraries-with-files
git switch add-test-libraries-with-files
```

.eslintrc.js
```diff
@@ -40,4 +41,15 @@ module.exports = {
         'prefer-const': ['error'],
         'semi': ['error', 'always'],
     },
+    overrides: [
+        {
+            files: [
+                '**/*.test.js',
```
DevJackSmith, line 47: tests should have their own folder in my opinion.

DevJackSmith, line 47:
```suggestion
                'tests/*.test.js',
```
I think we should give tests their own folder. The reason i think we should keep .test.js is because we might need supporting files there too.

tsconfig.json
```diff
@@ -1,26 +1,26 @@
...
+    "strictNullChecks": true,
+    "allowJs": true,
+    "paths": {
+      "@scripts/*":["src/scripts/*"]
+    }
   },
-  "include": [
-    "src/**/*"
-  ]
+  "include": ["src/**/*", "test/**/*"]
```
DevJackSmith, line 25:
```suggestion
  "include": ["src/**/*", "tests/**/*"]
```

hymccord:
Wasn't sure if preferred convention for test will spec `.test.js` or `.spec.js`. So I changed eslint to pick up both.

DevJackSmith (approved):

PR #244, opened 2022-11-12
Add Console Logging Util

hymccord:
- Add console logging util
- Add tests for console logger


DevJackSmith (approved):
clean code, with test, might not need it at the moment, but nice to have anyway.

Issue #240, opened 2022-11-07
Environment (localhost only) urls being set incorrectly


hymccord:
A bug was introduced from #239 for locally developed versions.

This initially sets all the URL to production location.
```javascript
    let base_domain_url = "https://www.mhct.win";
    const db_url = base_domain_url + "/intake.php";
    const map_intake_url = base_domain_url + "/map_intake.php";
    const convertible_intake_url = base_domain_url + "/convertible_intake.php";
    const map_helper_url = base_domain_url + "/maphelper.php";
    const rh_intake_url = base_domain_url + "/rh_intake.php";
```
If you are running version 0, the `base_domain_url` is changed to localhost.
However, this does not update all the other urls besides `base_domain_url` because they are all `const`.

If you are running the mhct website docker locally, this will cause hunts to retrieve a uuid from localhost (`http://localhost/uuid.php`). It will send a real message to prod server (`https://www.mhct.win/intake.php`). It will be dumped by the server since that uuid wasn't generated by redis (I think).

Also, not a bug, but the configuration for the local server could ideally be adjusted without it being detected by git. I run the mhct docker on a non-standard port since port 80/443 are already in use. You can use webpack to replace markers in the code at compilation.

PR #246, opened 2022-11-14
Stop sending user_id from extension in favor of hunter_id_hash

DevJackSmith:
- also cleaned up a lot of local test vars to align with testing full stack locally


hymccord:
Relying on the hash to be initialized really helped clean up those function calls. Nice work. Nothing stood out to me.

hymccord:
Oh, this fixes #240 so you might want to link that in the description

Issue #256, opened 2022-11-29
Timer no longer works due to new horn


hymccord:
HG has changed the html elements for the horn. 

Previous
```html
<div class="mousehuntHud-hornTimer">
  <span id="huntTimer">0:05</span>
</div>
```

New (abbreviated)
```html
<div class="huntersHornView">
  <div class="huntersHornView__timer huntersHornView__timer--default countdown">
    <div class="huntersHornView__timerState huntersHornView__timerState--type-countdown huntersHornView__countdown">0:05</div>
    <!-- this one is disabled til ready (comment by me) -->
    <div class="huntersHornView__timerState huntersHornView__timerState--type-ready">Ready</div>
  </div>
</div>
```

Relevant app.js functions that hg uses:

```js
_container = $('.huntersHornView');

  function setTimerCountdownSeconds(totalSeconds) {
    const mins = Math.floor(totalSeconds / 60);
    const seconds = totalSeconds % 60;
    const value = mins + ':' + (seconds < 10 ? '0' + seconds : seconds);
    $('.huntersHornView__countdown', _container).text(value);
  }

  // This fn is called with either setTimerState('ready') or setTimerState('countdown')
  function setTimerState(value) {
    $('.huntersHornView__timer', _container).removeClass('ready countdown').addClass(value);
  }
```

AardWolf:
ok, so if it hasClass('ready') it's ready for a honking -- appropriate for main.js line 135

Now to find the element we're clicking for line 139 and 141.

PR #257, opened 2022-11-29
🎺  New horn updates

bradp:
See #256

**Updates**
- Correctly counts down & checks for ready state
- Correctly sounds horn from alert
- Check for king's reward (still needs testing)

### DIFF
--- a/src/scripts/content.js
+++ b/src/scripts/content.js
@@ -56,13 +56,18 @@ chrome.runtime.onMessage.addListener((request, sender, sendResponse) => {
         let message = "Logged out";
         const krHudElement = document.getElementsByClassName('mousehuntHud-huntersHorn-response')[0];
         const krPageElement = document.getElementsByClassName('mousehuntPage-puzzle-form-state hasPuzzle')[0];
-        const hunt_timer = document.getElementById('huntTimer');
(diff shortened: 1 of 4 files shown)


src/scripts/content.js
```diff
@@ -56,14 +56,19 @@ chrome.runtime.onMessage.addListener((request, sender, sendResponse) => {
         let message = "Logged out";
         const krHudElement = document.getElementsByClassName('mousehuntHud-huntersHorn-response')[0];
         const krPageElement = document.getElementsByClassName('mousehuntPage-puzzle-form-state hasPuzzle')[0];
-        const hunt_timer = document.getElementById('huntTimer');
+        const hunt_timer = document.querySelector(".huntersHornView__countdown");
+        const is_ready = document.querySelector('.huntersHornView__timerState--type-ready');
```
AardWolf, line 60: I think we can just confirm the hunt_timer hasClass('ready')

bradp:
Pushed some changes for it. 

There's room for improvement, but currently working for me as far as I can tell. 👍

src/scripts/content.js
bradp: Updated in `7993742bcb21` 👍

AardWolf (approved):
I've been running with it locally for a while and it appears to work.

PR #258, opened 2022-11-29
Add utility class to interact with new horn

hymccord:
- Add utility class to interact with horn hud
- Use horn utility clas
- Add tests with DOM

Closes: #256

### DIFF
--- a/src/scripts/content.js
+++ b/src/scripts/content.js
@@ -1,3 +1,5 @@
+import {HornHud} from './util/HornHud';
+
 (function () {
 if (document.body == null) {
     return;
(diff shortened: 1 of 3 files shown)


hymccord:
One of the nice things I fixed was using this code:

```ts
// Since this is delayed by a few seconds from background script
// Check that user didn't beat us
if (!HornHud.isHornReady()) {
    return;
}
```

If you are paying attention to the page, you can spam the 'h' key or click the horn as soon as it arrives and the `confirm` web prompt won't happen (if you have that enabled). The unobstrusive alert will still happen; Aard has stated it would be nice to detect this as well but that is a more involved fix.

hymccord:
I have removed the tests as they are all based on the MH DOM which we were politely asked not to include.

I rebased and forced pushed so all of the offending code is not in the history.

AardWolf (approved):

PR #266, opened 2022-12-07
Add small delay between horn mouse down and up

hymccord:
After release, it was found that the horn wasn't clicking in the way when run in debug mode.

Use the same delay as bradp, but in an async/await way


AardWolf (dismissed):
I'm a fan of await!

AardWolf (approved):
now it has the mouseup

PR #269, opened 2022-12-13
Add initial hg typings

hymccord:
Add initial typescript typings that will be used for other PRs.

I expect these types to change a lot as we discover what can and can't be null or undefined from the hg api responses.

### DIFF
--- a/src/scripts/types/hg.ts
+++ b/src/scripts/types/hg.ts
@@ -1,0 +1,69 @@
+
+export interface HgResponse {
+    user: User;
+    page?: unknown;
+    success: 0 | 1;
+    active_turn?: boolean;
+}
+
+export interface User {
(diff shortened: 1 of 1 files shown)


src/scripts/types/hg.ts
```diff
@@ -0,0 +1,69 @@
...
+    bait_item_id: number;
+    trap_power: number;
+    trap_power_bonus: number;
+    trap_luck: number;
+    trap_attraction_bonus: number;
+    has_shield: boolean;
+    environment_name: string;
+    environment_id: number;
+    quests: Quests;
+    enviroment_atts: EnvironmentAttributes;
```
AardWolf, line 30: The theory is that when they fix this typo it would go away...

PR #271, opened 2022-12-16
Adding Dark Mode by PotatoSalad

DevJackSmith:
- added webpack action to fetch latest css files from PotatoSalad repo with their permission https://github.com/MHCommunity/mh-dark-mode
- added toggle in the options and loader in the context.js
- also moved tsitus autoloader files from github action to webpack

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -25,25 +25,6 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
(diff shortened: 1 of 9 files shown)


hymccord:
We need to do this for tsitu scripts as well! Very cool webpack plugin.

AardWolf (approved):
Huh. I was worried when all the tsitu ones were removed but I see where they went now.

PR #301, opened 2023-02-17
Move SB+ factory to separate file (as javascript)

hymccord:


AardWolf (approved):

AardWolf:
It'll be good to adjust the list of cheeses with IDs and just use the inventory part of the response (if it exists) like we do for the other non-convertible convertibles (like torch charms). Will still have to hardcode the snack packs, I think.

PR #305, opened 2023-02-21
Convert SEH to detailer module

hymccord:
* Started populating JournalMarkup type (other detailers will need this)
* Moved SEH to separate file. First example of a global detailer.

Removed the `calcEggHuntDetails` and converted that into an `IDetailer` conforming class.  
The `SpringEggHuntDetailer` is added to the `globalDetailersModules` array which is imported by the main script. There is no loss of functionality.


hymccord:
It's just the process of converting all the details over from main.js without changing the logic. The halloween details are still for cannon firing.

AardWolf (approved):

PR #306, opened 2023-02-27
Add Story Seeds snack pack to sb factory

hymccord:
Using the predicted item name by following the convention of previous items.


src/scripts/modules/ajax-handlers/sbFactory.js
```diff
@@ -142,6 +142,13 @@ class SBFactoryAjaxHandler extends AjaxSuccessHandler {
             'Rift Combat Cheese': 2096,
             'Null Onyx Gorgonzola': 2100,
             'Extra Rich Cloud Cheesecake': 3274,
+            'Mythical Mulch': 3451,
```
AardWolf, line 145: I would really like to get away from listing out all the cheesy outcomes but that does not have to be in this release.

AardWolf (dismissed):

Issue #308, opened 2023-02-28
Add Stager for SB Factory


AardWolf:
This is an opportunity to use the new stagers! We could add stages for the SB Factory using a mix of the existing logic and adding in logic for Factory Repair Charms (FRC). This would mean we'd need a "+ Repair" for each stage. For clarity we might need something to indicate the stage is without repair charm hunts when appropriate.

tlarkin011:
I would love to help with this issue!

PR #309, opened 2023-02-28
SB+ Factory Stager

hymccord:
- Move SBFactory to stager module
- Add tests for sb factory stager


AardWolf (approved):

AardWolf:
Partially implements #308 - whether factory repair charms need to be considered is up for debate.

PR #316, opened 2023-03-09
Allow clicking through the popup messages

bradp:


DevJackSmith (approved):

DevJackSmith (approved):
I would suggest making the notification smaller, but this is not exclusive to the other solution

PR #327, opened 2023-03-25
Fixing release notes generation

DevJackSmith:
the step we used didn't work, so trying out the one that's built into release generator. In my tests it worked great, but will have to see in main branch.


hymccord:
Default release notes should be good. 

If we want to tweak further
https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes

hymccord (approved):

PR #340, opened 2023-04-11
Add animation check to see if horn can be clicked

hymccord:
This will fix the horn alert (the intrusive one) popping up even though the horn is currently being sounded.

Previously, if the horn was sounding (animating) it would still have the `ready` class. Now it will prevent the popup if it's `ready` or animating.


DevJackSmith (approved):

DevJackSmith:
lgtm, thanks

DevJackSmith (approved):
lgtm. Warnings about "type = any" are unrelated and can be fixed in a different pr.

PR #346, opened 2023-04-21
Rewrite save-remote-file-webpack-plugin

hymccord:


hymccord:
Using `got` to nab the files instead of `download` (the vulnerable package).

However, got v12 only EmcaScript Modules which I was unable to get working in a short time. Maybe something to look at later.
Could use `got` v11, but `got-cjs` is published as well but just as a CommonJS module.

DevJackSmith (approved):

PR #369, opened 2023-05-14
Add Bountiful Beanstalk castle room name to stage

hymccord:
- Add current room to BB stage
- Update BB filter exemptions due to new stage names
- Update exemption tests to use stager


hymccord:
Easiest to view by commit.

Last commit refactors the exemption tests to not hardcode the BB stage names into the tests and use the stager instead. Previously the tests would keep passing if you changed the stage names in the staging module. Now they will fail if you don't update the exemptions when you update the stager.

AardWolf (approved):

PR #393, opened 2023-06-11
Added [MHCT Scav Helper (auto)] Button

u-nel:
Additional button for Scavenger Hunt Helper.

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -13,7 +13,7 @@ import * as detailingFuncs from './modules/details/legacy';
     'use strict';
 
     let base_domain_url = "https://www.mhct.win";
-    let main_intake_url, map_intake_url, convertible_intake_url, map_helper_url, rh_intake_url, rejection_intake_url;
+    let main_intake_url, map_intake_url, convertible_intake_url, map_helper_url, rh_intake_url, rejection_intake_url, scav_helper_url;
 
     let mhhh_version = 0;
     let hunter_id_hash = '0';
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -249,6 +256,32 @@ import * as detailingFuncs from './modules/details/legacy';
...
+    }
+
+
+    function getTreasureClue(data, uncaught_only, input_name) {
+        if(input_name == 'mice')
+            return getMapMice(data, uncaught_only);
+        if(input_name == 'items')
+            return getMapLoot(data, uncaught_only);
+    }
+
```
hymccord, line 284: Since this function is a duplicate of the other, we should simplify both by accessing the property based on what type of map it is.

```javascript
    /**
     * Extract goals from map (can be scav or treasure)
     * @param {Object} data Data response from managers/ajax/users/treasuremap.php
     * @param {boolean} remaining_only Boolean, if true, to filter out completed goals
     * @returns {string[]} Goal names. If on treasure map, it will be mice names and if
     * on scavenger it will be item names.
     */
    function getMapGoals(data, remaining_only) {
        const goalCategory = data.treasure_map.is_scavenger_hunt ? 'item' : 'mouse';
        const goals = {};
        $.each(data.treasure_map.goals[goalCategory], (key, goal) => {
            goals[goal.unique_id] = goal.name;
        });

        if (remaining_only) {
            $.each(data.treasure_map.hunters, (key, hunter) => {
                $.each(hunter.completed_goal_ids[goalCategory], (key, goal_id) => {
                    delete goals[goal_id];
                });
            });
        }

        return Object.values(goals);
    }
```

hymccord (changes requested):
The logic might need some rearranging if others agree it should be one button.

DevJackSmith:
Agree, we should use one button and reuse most of existing code. In fact everything might already work, we need to test

u-nel:
I will try to submit again. This was like a trial for me and rushly developed.
Thank you for the feedbacks.

hymccord:
In the future, keep the same PR. It allows us to see that all feedback is addressed as the new PR won't have our previous comments.

PR #398, opened 2023-06-13
Support: Scavenger Hunt on MHCT Map Helper (auto)

u-nel:
Hello, 
Implemented the suggested flow and code.

Thank you.


src/scripts/main.js
```diff
@@ -192,61 +193,80 @@ import * as detailingFuncs from './modules/details/legacy';
...
+        treasure_map = treasure_map.filter(map => ['treasure', 'event'].includes(map.map_class));
+
+        if(!treasure_map.length) {
+            alert('This seems to be a new kind of map and not yet supported.');
             return;
         }
 
         const payload = {
-            map_id: user.quests.QuestRelicHunter.default_map_id,
+            map_id: treasure_map[0].map_id,
```
hymccord, line 214: Does anyone know what happens if you are on a Treasure/Scav map at the same time as an special event map (e.g. Spooky, GWH). Unfortunately, I don't have one of these scrolls on me at the time to test.

I believe the `default_map_id` always returns the Treasure/Scav map id.

u-nel: I tried Bounty map .. If you are in Claw Shot City .. Bounty map is default

hymccord: I was testing around. The `default_map_id`, will change depending on location. So the problem I see now is if you are on a Treasure + Event map, this may never help out with the other given the arrangement of the `maps` array.

You are on both a Treasure Map (e.g. Slayer) + Event Map (e.g. GWH) and in the GWH area, if the maps array has the slayer map first you won't ever get the help for the event map. The reverse could also happen where you never get slayer help if the GWH map is first.

u-nel: The solution I see for that is to loop through the map array and open two windows of the helper.

PR #399, opened 2023-06-14
Add new BB SB+ mouse to stage exception rule

hymccord:
Allow recording transition to Beanstalk boss when catching the new SB|+ mouse.


DevJackSmith (approved):
Lgtm

PR #405, opened 2023-06-21
Add Bounty Hunter exemption

hymccord:
- Add Claw Shot City quest type
- Add bounty hunter reject test case
- Add stage exemption for Bounty Hunter catch


hymccord:
Straightforward. If Bounty Hunter is caught without a map and causes a stage transition: allow it.

By commit:
Wrote quest type first.  
Next, a test to assure rejection.  
Then wrote the exemption so the test needed to changed to accept it.

DevJackSmith (approved):

PR #453, opened 2023-09-14
Update KGA for 2023

hymccord:
Migrating KGA to an ajax handler, fix a bug I found in the recordPrizePack code and add a util method to convert formatted strings (`5,000`) to numbers.

Pushing batches of commits so CI will be green

- [x] Move to separate file so it can be tested
- [x] Write tests (found bug)
- [x] Migrate against existing tests
- [x] Add king's vault as convertible (id: `130009`).


DevJackSmith (approved):
lgtm

PR #455, opened 2023-09-16
Automatic version allow

DevJackSmith:
- simplified date formatting
- added ssh to server step where it checks if version is already allowed, and if not, allows it
- created a new key pair just for github actions and set server connection info in github secrets

tested and it works

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -49,6 +49,7 @@ jobs:
           echo "ext-version=$EXT_VERSION" >> $GITHUB_OUTPUT
           # Remove dots and pad to two digits and remove spaces
           EXT_VERSION_NUM=$(echo $EXT_VERSION | sed 's/\./ /g' | xargs -n1 | awk '{printf "%02d", $1;}' | xargs)
+          echo "Setting extension version number to $EXT_VERSION_NUM"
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -37,127 +38,157 @@ jobs:
...
-        run: |
-          git tag ${{ env.EXT_VERSION }}
-          git push --tags
-
-  chrome-submit:
-    name: Submit to Chrome
+          echo "Setting extension version to $EXT_VERSION"
+          echo "ext-version=$EXT_VERSION" >> $GITHUB_OUTPUT
+          # Remove dots and pad to two digits and remove spaces
+          EXT_VERSION_NUM=$(echo $EXT_VERSION | sed 's/\./ /g' | xargs -n1 | awk '{printf "%02d", $1;}' | xargs)
```
hymccord, line 52: It may be easier to understand to just use `date` again and supply a new format string.

With `+%y.%-m.%-d` (from above), `-` (hyphen) says do not pad the field

So you can just do `date +%y%m%d ` and get `230916`

DevJackSmith: this is not always date. this is sometimes custom date, sometimes date + 1

hymccord: Oh true. Good catch.

hymccord (approved):

PR #508, opened 2023-12-17
testing found bugs!

AardWolf:


tests/scripts/hunt-filter/exemptions/environments/iceFortress.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+        beforeEach(() => {
+            preUser = {...getDefaultUser(), ...getIceFortressUser()};
+            postUser = {...getDefaultUser(), ...getIceFortressUser()};
+            preMessage = {...getDefaultIntakeMessage()};
+            postMessage = {...getDefaultIntakeMessage()};
+        });
+
+        it('should accept when no transitions', () => {
+            preUser.quests.QuestIceFortress = {shield : {is_broken: false},
+                cannons:    {
```
hymccord, line 36: Since these are all the same in all the tests, I would just pull this into the default `getIceFortressUser()` then change what is needed. Which is usually nothing in these cases.

PR #528, opened 2024-02-04
Updating github actions to v4

logicalup:
not sure how we can test other than trying to release


hymccord (approved):

hymccord (changes requested):
I think a few more need to be updated

`actions\setup-node@v4`

AardWolf (approved):

hymccord (approved):

PR #535, opened 2024-02-09
Clarify horn readiness option's descriptions

hymccord:
Even as a developer who is intimate with the code, I can _never_ remember what each of these option do and don't do.

Suggestions on wording are welcome.


tehhowch (commented):
Seems maybe more appropriate to provide tooltips (via HTML `title` attribute) that offer more info, or use accordions (via `<details>` tags) to show examples of each.

AardWolf (approved):

Issue #211, opened 2022-10-07
GES Stage Update


AardWolf:
Supply Depot - No Rush should have another stage to indicate the Supply Schedule Charm is armed.

Issue #169, opened 2022-04-15
Gnawnia Express Station Stages Broken


AardWolf:
Possibly never worked. Issue discussion got re-picked up here: https://discord.com/channels/275500976662773761/355474934601875457/964516539510431814

Short form is this stage uses message.charm.id, does a === compare between a string and number, fails.

Should probably compare with user.trinket_item_id but the message object is likely already populated. Should probably do an == compare or parsed to an integer for the compare.

PR #556, opened 2024-05-13
Convert GES to stager

hymccord:
- Add Gnawnian Express Station stage tests
- Add Gnawnian Express Station quest type
- Convert GES to stager and remove obsolete legacy file
- Convert string ids into numbers during hunt parse

Much of the debug logging removed here is redundant. Any traps changes will be caught by rejection engine.

Fixes: #169

Close: #211 (already done previously)


hymccord:
Last of legacy staging code done!

AardWolf (approved):

PR #569, opened 2024-07-13
Add HUD Message integration

hymccord:
- Add new option to support HUD messages
- Integrate flash messages with HUD


hymccord:
Three options: HUD (default), Banner (current), Toast.

Mockup: 
image

Toast is same style as banner but just bottom right. Similar to jQuery toasts.

AardWolf (approved):

PR #574, opened 2024-08-07
Restructure type files

hymccord:
Preparing for my [zod](https://zod.dev/?id=introduction) work to be introduced.

Summary:

Break apart `types\hg.ts` into a `types\hg` folder. Move quest type files into that folder.


src/scripts/types/hg/inventoryItem.ts
```diff
@@ -0,0 +1,14 @@
...
+    /** HitGrab's internal number id of item */
+    item_id: number;
+    /** Friendly display name of item */
+    name: string;
+    /** Unique snake_case identifying name of item */
+    type: string;
+    /** Item category: bait, crafting, stat, etc... */
+    // classification: Classification;
+    /** Total amount of item in user inventory */
+    quantity: number;
```
AardWolf, line 13: sometimes it has a comma and is in quotes - this forces it to turn into a number?

AardWolf (approved):

src/scripts/types/hg/inventoryItem.ts
hymccord: This is how we currently view JSON responses from them. We just assume it's a number, but we're going to soon verify that with the Zod library.

PR #576, opened 2024-08-15
Add support for Draconic Depths

hymccord:
- Add quest type for Draconic Depths
- Add stages for Draconic Depths
- Add transition exemptions for Draconic Depths


AardWolf (approved):
Approved with a small comment

AardWolf (approved):

PR #584, opened 2024-09-12
Fix sticky HUD messages

hymccord:
- **Remove HornHud undefined function call**
- **Use queue for multiple hud messages**

Multiple calls of `HornHud.showMessage` in a row would result in the previous message coming back and getting stuck. Implemented a queue system to process one message at a time.

### DIFF
--- a/src/scripts/util/hornHud.ts
+++ b/src/scripts/util/hornHud.ts
@@ -3,10 +3,34 @@
  * This class can only be uses from contexts that have access to DOM.
  */
 export class HornHud {
+    private static messageQueue: { message: string, type: 'success' | 'warning' | 'error' }[] = [];
+    private static isProcessingQueue = false;
 
     public static async showMessage(
         message: string,
         type: 'success' | 'warning' | 'error' = 'success'
+    ) {
+        this.messageQueue.push({message, type});
+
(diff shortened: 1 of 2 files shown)


src/scripts/util/hornHud.ts
```diff
@@ -3,10 +3,34 @@
...
+    ) {
+        this.messageQueue.push({message, type});
+
+        if (!this.isProcessingQueue) {
+            await this.processQueue();
+        }
+    }
+
+    private static async processQueue() {
+        this.isProcessingQueue = true;
```
AardWolf, line 21: I don't know JS async well enough but other multi-threaded processes could still have a race condition handling flags this way. Does JS have a "mutex" concept? Or does it turn these things into a mutex-like thing?

Otherwise, to harden it you'd be setting it to something like a thread-internal identifier and then confirming it is set to your thread identifier before proceeding.

AardWolf (approved):

src/scripts/util/hornHud.ts
hymccord: Not in the case for browsers. Javascript is always single-threaded here.... as long as we don't use Web Workers.

PR #587, opened 2024-09-24
Introduce Zod library for verifying JSON

hymccord:
- Add zod as dependency
- Convert all hg types to use zod
- Fix errors related to stricter types
- Warn when encountering unexpected responses

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,6 +8,9 @@
       "name": "mhct-extension-env",
       "version": "1.0.0",
       "license": "MIT",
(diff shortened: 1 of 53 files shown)


hymccord:
Taking each type file and converting it to it's respective Zod shape and exporting the type like previous.
Relying on all the previously built tests to ensure everything still works as expected.

Right now, no hunts will be affected. All I do is log a warning if unexpected types are encountered in main.js.

AardWolf (approved):

Issue #628, opened 2025-04-15
Relic Hunter hints not being submitted


AardWolf:
Looks like we missed a spot for the change in treasure map endpoints: https://github.com/m-h-c-t/mh-helper-extension/blob/978785dc80ebc8cab8082637dc0dd488b4ad6ad6/src/scripts/main.js#L351

PR #629, opened 2025-04-15
Updated endpoint for RH hints

AardWolf:
Relic hunter hints were not being submitted because the _v2 endpoint is now being used in-game.


AardWolf:
Addresses issue #628

PR #632, opened 2025-05-01
Validated Ajax Handlers w/ Treasure Map Example

hymccord:
- **Add new base class for zod validated ajax success handlers**
- **Convert recordMap to new validated ajax success handler**

`AjaxSuccessHandlers` can now inherit from `ValidatedAjaxSuccessHandler` and define a Zod schema in a field. That schema will be used and if it matches will be pass down to a `validatedExecute` method. The data is guaranteed to match the defined schema! No more needing to assume data type.

This will eventually lead to the removal of event date flags!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,6 +26,7 @@
         "html-webpack-plugin": "^5.6.3",
         "jest": "^29.7.0",
         "jest-environment-jsdom": "^29.7.0",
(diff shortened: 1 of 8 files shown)


hymccord:
I plan on converting the other handlers to the validated ones after this base class is merged.

AardWolf (approved):

PR #634, opened 2025-05-03
KGA 2025

hymccord:
* Migrate KGA to validated ajax handler

Can now remove time gate checks due to guaranteed JSON structure.
Removed `parseHgInt` and added new `zodStrumber` to help define string like numbers with commas. (HG *sometimes* returns numbers with commas... but not most of the time)


hymccord:
One change to notice: instead of hard-coding the items that are in the vault, it will fetch (and cache) all items from `api.mouse.rip`.

AardWolf (approved):
"Strumber"

PR #636, opened 2025-05-06
Add abstraction for mouse.rip API. Fix KGA handling.

hymccord:
Instead of using ApiService directly, it's smarter to create a wrapper for each website we reach out to. Exactly how SubmissionService uses ApiService and is exclusively for mhct.win.

KGA handling had two problems: one was accidentally sending a body on a GET request which isn't allowed in browsers and a zod related error was that the item property `item_id` was actually just `id`. 

I've fixed the docs on api.mouse.rip to reflect what the api actually returns.


hymccord:
Waiting for chrome review ⏲️

AardWolf (approved):

PR #648, opened 2025-09-04
Add support for Afterword Acres & Epilogue Falls

hymccord:
* **Add minimal quest schemas for AA and EF**
* **Add Afterword Acres stages for blight tiers**
* **Add stage transition exemption for AA blight**
* **Add detailer for Epilogue Falls zone**
* **Add details for Rainbow Quill procs**


hymccord:
Wow, I thought it was Afterwood....

After**word**. Fixing.

Issue #57, opened 2017-11-14
Record active auras


groupsky:
Some auras affect loot drops or catch rates, so it's best if all active auras are recorded with each hunt.

hymccord:
Related mh-hunt-helper: https://github.com/m-h-c-t/mh-hunt-helper/issues/222

PR #651, opened 2025-09-20
Add aura tracking

hymccord:
This will require a change on the hunt helper side

Related: https://github.com/m-h-c-t/mh-hunt-helper/issues/222
Closes: #57


hymccord:
image

